PR #45, opened 2015-02-10
non-clickable link for {URL} tag -> clickable one

zdzis:
I have checked on bn, ca, da, de, el, en, et, fa, fr, hy, is, it, la, nl, no, pl, pt, ro, ru, te, vec and old wikisources. Only on the plwiki the {URL} tag is used. Changing into a clickable link may increase its popularity :), and enhance “About” section usability.


Tpt:
Thank you!

PR #54, opened 2015-05-31
Use Bootstrap from cdnjs

ricordisamoa:
Lots of libraries are now mirrored on Tool Labs:
https://lists.wikimedia.org/pipermail/labs-l/2015-April/003670.html


Tpt:
Thank you!

PR #71, opened 2016-02-09
cover thumb picture for oldwikisource local file

zdzis:
generate cover thumb picture if 'ws-cover' is a local oldwikisource file - avoid `cURL error 3: <url> malformed` error, see: https://tools.wmflabs.org/wsexport/tool/book.php?lang=www&page=User%3AZdzislaw%2Ftest%C5%9B%C4%87+P02&format=epub-3


Tpt:
Thank you!

PR #73, opened 2016-02-16
Fix grammar

jberkel:
See discussion on en.wikisource.org

https://en.wikisource.org/wiki/Wikisource:Scriptorium#Kindle_users:_Please_check_mobi_export_form

### DIFF
--- a/book/i18n.ini
+++ b/book/i18n.ini
@@ -1,5 +1,5 @@
 title_page = "Title page"
 cover = "Cover"
-exported_from_wikisource_the = "Exported from Wikisource the %d"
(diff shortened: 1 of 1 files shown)


zdzis:
renaming the key from _`exported_from_wikisource_the`_ into _`exported_from_wikisource`_ is not a good idea - changes will be required in all ws local [MediaWiki:Wsexport i18n.ini](https://pl.wikisource.org/wiki/MediaWiki:Wsexport_i18n.ini) files.

jberkel:
ok, wasn't aware of this, changed

Tpt:
Thank you very much!

PR #74, opened 2016-02-17
CLI: make credit generation optional

jberkel:
Adding credits can add a lot of time to the book generation
  process (in the order of minutes) and is not always desirable, e.g. when experimenting
  with output formats or working on the tool itself.


Tpt:
Thank you!

PR #76, opened 2016-02-18
Language needs to get passed through

jberkel:


jberkel:
I'd be better to write a test for this, which depends on #77

Tpt:
Thank you! Feel free to write the test in an other change.

PR #81, opened 2016-03-14
Run epubcheck and display warnings

jberkel:
This PR integrates [epubcheck](https://github.com/IDPF/epubcheck) into the build pipeline. Errors flagged by epubcheck currently don't fail the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ dist: trusty
 
 before_install:
   - sudo apt-get update -qq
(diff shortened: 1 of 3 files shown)


jberkel:
I'm wondering if the 503s are caused by rate limiting?

Tpt:
> I'm wondering if the 503s are caused by rate limiting?
It's definitely possible. But as we do not usually encounter this issue during real wsexport use it could just be a punctual problem on Wikimedia servers. The fact that the build pass now suggests the later option.

PR #82, opened 2016-03-14
composer.lock should be checked in

jberkel:
> Commit your application's composer.lock (along with composer.json) into version control.
This is important because the install command checks if a lock file is present, and if it is, it downloads the versions specified there (regardless of what composer.json says).

https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file

 - use makefile to install composer & run composer tests
 - set executable bit on book.php
 - update & reflow README


ricordisamoa:
MediaWiki for example leaves autogenerated files like composer.lock out of version control.

jberkel:
Is there a reason for this though? Most similar systems (rubygems, cocoapods etc.) all recommend checking in lock files, to have consistent & reproducible builds.

samwilson:
MediaWiki is a bit annoying in its handling of composer.json I reckon: they should either ship it with composer.lock, or change composer.json to composer.example.json. Because now Composer is more and more often the means of managing extension installation, a site's composer.json is customised, and so shouldn't be in version control.

Sorry... off topic. All I wanted to say was yeah, good plan to check in composer.lock. :-) It's not an installation-specific artefact in this instance.

jberkel:
@Tpt does this PR need anything else to get merged?

PR #93, opened 2016-04-11
Update composer dependencies

jberkel:


Tpt:
I'm not sure it's a good idea to commit the composer.lock file because it would make wsexport compatible with PHP 5.6+ even if the WIkimedia Tools labs were it's deployed is still using PHP 5.5. More, it's not in Wikimedia practices to commit the composer.lock files.

So, I'm against this change if a good solution is not found

jberkel:
hmm, i suggest we set up travis in a way that all tests get run against php 5.5 (or whatever is deployed on labs) and 5.6, to ensure nothing breaks there and in future updates.

Tpt:
Travis now runs against PHP 5.5, 5.6, 7 and HHVM

I close this PR for now

jberkel:
nice work! still don't understand why a checked in composer.lock would prevent the tests / code run on 5.5 though?

Tpt:
> nice work! still don't understand why a checked in composer.lock would prevent the tests / code run on 5.5 though?
Yes, we could create a checked in composer.lock with the last version of dependences that would work with PHP 5.5. But it would stop us to use newer version of dependences like PHPUnit 5 on PHP 5.6+

Issue #89, opened 2016-03-17
Missing content in some ebooks


jberkel:
See for example https://en.wikisource.org/wiki/Tales_of_Unrest

The chapter picking logic (https://github.com/wsexport/tool/blob/master/book/BookProvider.php#L468) fails in this particular case since not all entries are under the /Book_Title/ url namespace. An option would be to fix the data but I suspect that's not always possible + would invalidate existing urls.

Another option would be to follow the link and then check the metadata of the target page (e.g. `{{header previous = ... }}`). Any thoughts?

zdzis:
hello,
@jberkel there is no problem in wsexport...
the chapter picking logic is fixed for years to collect chapters:
- are subpages under the /Book_Title/, OR
- links (sub-chapters) in the wikitext are inside a tag with id = "ws-summary" [BookProvider.php#L443](https://github.com/wsexport/tool/blob/master/book/BookProvider.php#L443) (see: [wsexport microformat](https://wikisource.org/wiki/Wikisource:Microformat) )

If there are links that should be taken as the book sub-chapters (but they are not sub-pages under the /Book_Title/) they should be just covered by the tag with proper id, such as: 
```
<span id="ws-summary">
*[[Karain]]
*[[The Idiots]]
*[[An Outpost of Progress]]
*[[The Return (Conrad)|The Return]]
*[[The Lagoon]]
</span>
```

So, changes on the https://en.wikisource.org/wiki/Tales_of_Unrest are required, not in wsexport tool.
Besides, the change of the current chapter picking logic would require changes to the hundreds of thousands current pages on ws, which are created using a current logic and microformat.

Z.

jberkel:
thanks, wasn't aware of the microformat, it solves the problem nicely. except it's hard to say how many books still need "fixing".

jberkel:
Just a minor correction: it needs to be `<div id="ws-summary>`, not `<span>`. With these changes I successfully exported the whole book, closing this one.

PR #98, opened 2016-04-17
Page parser test

jberkel:
Add tests for some behaviour described in #89


.travis.yml
```diff
@@ -17,7 +17,9 @@ before_install:
 install:
   - make install
 
+before_script:
```
Tpt: Why? composer-tests are tests just like the others.

jberkel: right now they are not much more than a linter step, which are normally run before tests. this way simple code style problems will fail the build immediately without running all the other (slow) tests. but happy to move this out of the PR.

Tpt: I see, good point. So, I'm ok with it

PR #118, opened 2016-09-27
include fonts in .css (even when not embedded in ePub [fixes #26])

ousia:
@Tpt, this pull request adds typefaces in the stylesheet file, as described the issue.

Having the typefaces even when not embedded will help, if the user has installed on the system.


ousia:
@Tpt, I’m closing the pull request myself, since my approach is too intrusive.

PR #119, opened 2016-10-16
add FreeSerif and LinuxLibertine (if and only if previously defined)

ousia:
@Tpt, this is a better way to have the fonts in the ePub.

It doesn’t change the default typeface, since either `f_s` or `l_l` have to be defined first with `@font-face`.

Any user can make use of the typefaces defining them.

I hope this is better now and can be merged.

### DIFF
--- a/book/FontProvider.php
+++ b/book/FontProvider.php
@@ -15,11 +15,11 @@ class FontProvider {
 	 */
 	protected static $data = [
 		'freeserif' => [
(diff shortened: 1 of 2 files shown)


Tpt:
Thank you very much for finding this way. It's very nice! Names like "wsexport-LinuxLibertine" would probably be more readable.

ousia:
@Tpt, the new names are `wse_FreeSerif` and `wse_LinuxLibertine`.

I agree with you, they are now more understandable (or less cryptic :wink:).

PR #121, opened 2016-10-17
Update FontProvider.php

ousia:
We need a specific variable `css_names`

Otherwise the wrong font names will be chosen for font embedding (with no actual embedding). And `content.opf` will list the wrong font names too..

Sorry, because I didn’t realize that yesterday.


Tpt:
I definitely should have checked more carefuly...

PR #143, opened 2017-12-05
Check if filename is set

stephanedelprat:
`createEpub` can throw an exception, in this case, `$persistentEpubFileName` is not set, so `removeFile('')` is called and tries to remove the current directory. (yes, it fails because `rm` does not delete a directory, without `-Rf`)


Tpt:
Thank you for the change!

Issue #146, opened 2018-05-16
error: "Page revision not found"


ousia:
Trying to generate an ePub version with the Greek works of Plato throws an error message:

> Page revision not found.

The link reads <https://tools.wmflabs.org/wsexport/tool/book.php?lang=el&page=Συγγραφέας:Πλάτων&format=epub-3&fonts=>.

I wonder what triggers the issue.

samwilson:
I've attempted to fix this in #151.

Tpt:
It seems to be fixed thanks to @samwilson

ousia:
@samwilson, many thanks for your fix.

PR #151, opened 2019-03-22
Use 'href' instead of 'title' as the source of chapter links

samwilson:
Change the chapter-finding method to use the href value of
anchors rather than the title, because linked images can set
different title text.

Bug: #146

### DIFF
--- a/book/PageParser.php
+++ b/book/PageParser.php
@@ -40,17 +40,21 @@ class PageParser {
 	}
 
 	/**
-	 * return the list of the chapters with the summary if it exist.
+	 * Return the list of the chapters (based on the ws-summary HTML if it exists, otherwise via all internal links).
+	 * @param string[] $pageList Array of page titles.
+	 * @param string[] $namespaces Array of all localized namespace names of the current wiki.
 	 * @return Page[]
-	 * TODO retrive only main namespace pages ?
 	 */
 	public function getChaptersList( $pageList, $namespaces ) {
 		$list = $this->xPath->query( '//*[@id="ws-summary" or contains(@class,"ws-summary")]/descendant::a[not(contains(@href,"action=edit") or contains(@class,"extiw") or contains(@class,"external") or contains(@class,"internal") or contains(@class,"image"))]' );
 		$chapters = [];
 		/** @var DOMElement $link */
 		foreach ( $list as $link ) {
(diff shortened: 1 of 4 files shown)


samwilson:
I'm not sure this is completely correct. For example, what happens if a link's href has a query string component? (Not that I can think of why it would.)

book/PageParser.php
```diff
@@ -40,17 +40,20 @@ public function metadataIsSet( $id ) {
...
-	 * TODO retrive only main namespace pages ?
 	 */
 	public function getChaptersList( $pageList, $namespaces ) {
 		$list = $this->xPath->query( '//*[@id="ws-summary" or contains(@class,"ws-summary")]/descendant::a[not(contains(@href,"action=edit") or contains(@class,"extiw") or contains(@class,"external") or contains(@class,"internal") or contains(@class,"image"))]' );
 		$chapters = [];
 		/** @var DOMElement $link */
 		foreach ( $list as $link ) {
-			$title = str_replace( ' ', '_', $link->getAttribute( 'title' ) );
+			// Extract the page title (including namespace) from the relative URL in the link.
+			$title = urldecode( substr( $link->getAttribute( 'href' ), strlen( '/wiki/' ) ) );
```
Tpt: It would be nice to also remove the fragment (after the first '#')

`explode( '#', $url )[0]` should do the job.

samwilson: Good point. I've changed it to use `parse_url()`, and added some tests.

utils/Api.php
```diff
@@ -173,14 +173,14 @@ public function getPageAsync( $title ) {
...
 			if ( isset( $page['revisions'] ) ) {
 				foreach ( $page['revisions'] as $revision ) {
-					return getXhtmlFromContent( $this->lang, $revision['*'], $page['title'] );
+					return getXhtmlFromContent( $this->lang, $revision['*'], $title );
 				}
 			}
 		}
-
-		throw new HttpException( 'Page revision not found', 404 );
+		throw new HttpException( "Page revisions not found for: $title", 404 );
```
Tpt: The integration test fails because `BookTest::testGetNonExistingTitleDisplaysError` expects the error message to be "Page revision not found" with revision singular.

samwilson: Fixed.

Tpt:
Thanks! I just deployed it on labs

PR #154, opened 2019-03-27
Check for URL path component before using

samwilson:
Some pages were throwing a PHP Notice: "Undefined index: path"


samwilson:
The log seems to have a few of these notices, but I'm not sure what page is generating them.

Tpt:
Deployed

samwilson:
Thanks! :)

I'll keep an eye on the logs and see what else goes wonky.

Issue #145, opened 2018-03-26
Unable to use wsexport  from telugu wikisource due to 502  bad gateway error


arjunaraoc:
While trying to download a pdf book from Telugu wikisource, 502 Bad gateway error occurs everytime.

Example page: https://te.wikisource.org/wiki/%E0%B0%A8%E0%B0%BE_%E0%B0%9C%E0%B1%80%E0%B0%B5%E0%B0%BF%E0%B0%A4_%E0%B0%AF%E0%B0%BE%E0%B0%A4%E0%B1%8D%E0%B0%B0-1

Click PDF icon on the right, and the link https://tools.wmflabs.org/wsexport/tool/book.php?refresh=true&lang=te&format=pdf-a5&page=%E0%B0%A8%E0%B0%BE+%E0%B0%9C%E0%B1%80%E0%B0%B5%E0%B0%BF%E0%B0%A4+%E0%B0%AF%E0%B0%BE%E0%B0%A4%E0%B1%8D%E0%B0%B0-1 results in 502 Bad gateway error

Basic tool webpage is functional at the page https://tools.wmflabs.org/wsexport/tool/book.php
I got same error when I tried on English Wikisource as well.

balajijagadesh:
i was able to download in epub format in the link given above sometimes. But at other times getting 502/503 gateway error

arjunaraoc:
Thanks @balajijagadesh, I too am experiencing  similar result.

jmatazzoni:
[...]

PR #157, opened 2019-03-30
Adds Mukta font

### LINKED ISSUE #147: Font Selection during export
Can we have an option for font selection during book export?? The problem/issue is in Indic Font. For example for Gujarati Wikisource. The font used in conversion is [Samyak font](https://fedoraproject.org/wiki/Samyak_fonts) which is decorative font and not comfortable in reading. If we can use [Lohit Gujarati](https://github.com/mozilla-b2g/moztt/tree/master/Lohit) or similar font that will be better.

Tpt:
Closes #147


Tpt:
@Sushant-savla Sadly the server seems a bit overloaded at this moment. This problem should not be related with the fonts but seems to be the same as #145

Sushant-savla:
I used https://tools.wmflabs.org/wsexport/tool/book.php to convert book in PDF format. However Out put is in Lohit Font and combined letters are show as separate Letters (page 12)
[ત્રિશંકુ.pdf](https://github.com/wsexport/tool/files/3064946/default.pdf)
. Thus the intendened changes are not effected.

PR #158, opened 2019-04-15
Update Libertinus fonts

julroy67:
Update Libertinus fonts

6.4 ⇒ 6.8


samwilson:
Are these files from https://github.com/libertinus-fonts/libertinus/releases/tag/v6.8 ?

I think we should add some info about where they come from, and perhaps include the whole release here — or, alternatively, add an updating tool to the makefile and not include these assets in this repo at all.

julroy67:
Yes, that's right, I picked them up from there.

If you expect to add more fonts in future releases, the repo may grow in size and as the fonts are not under the same license as the other files of the repo it may be better to not include the fonts at all in the repo.

Issue #44, opened 2015-02-04
Failure when there are too many too big images


Tpt:
Example: https://fr.wikisource.org/wiki/Voyage_au_centre_de_la_Terre

jmatazzoni:
This is definitely something we will look into for CommTech's new project on [E-Book Export](https://meta.wikimedia.org/wiki/Community_Tech/Improve_export_of_electronic_books). For the record, though, I just downloaded this as EPUB and the book came out just fine. It appears the problems we see with export are intermittent.

MaxSem:
Appears to be an out-of memory condition:
```
2019-04-17 13:00:39: (mod_fastcgi.c.2543) FastCGI-stderr: PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 212992 bytes) in /mnt/nfs/labstore-secondary-tools-project/wsexport/tool/vendor/guzzlehttp/guzzle/src/Handler/CurlMultiHandler.php on line 108
2019-04-17 13:00:39: (mod_fastcgi.c.2543) FastCGI-stderr: PHP Stack trace:
2019-04-17 13:00:39: (mod_fastcgi.c.2543) FastCGI-stderr: PHP   1. {main}() /mnt/nfs/labstore-secondary-tools-project/wsexport/tool/http/book.php:0
2019-04-17 13:00:39: (mod_fastcgi.c.2543) FastCGI-stderr: PHP   2. BookCreator->create() /mnt/nfs/labstore-secondary-tools-project/wsexport/tool/http/book.php:38
2019-04-17 13:00:39: (mod_fastcgi.c.2543) FastCGI-stderr: PHP   3. BookProvider->get() /mnt/nfs/labstore-secondary-tools-project/wsexport/tool/book/BookCreator.php:32
2019-04-17 13:00:39: (mod_fastcgi.c.2543) FastCGI-stderr: PHP   4. BookProvider->getMetadata() /mnt/nfs/labstore-secondary-tools-project/wsexport/tool/book/BookProvider.php:40
2019-04-17 13:00:39: (mod_fastcgi.c.2543) FastCGI-stderr: PHP   5. BookProvider->getPicturesData() /mnt/nfs/labstore-secondary-tools-project/wsexport/tool/book/BookProvider.php:154
[...]

PR #160, opened 2019-04-20
Double PHP memory limit

MaxSem:
While it's not a proper fix for #44, it improves the situation while
we're investigating it further. In any case, 256M is not an unreasonable
limit.


Tpt:
Thanks!

PR #166, opened 2019-05-02
Move config to separate file

samwilson:
Move the $wsexportConfig global to a separate config.php file
and load it for the web and CLI entry points.

As this new file is not optional, also add a composer script to
create it at install and update time.

Bug: T222330


MaxSem:
Needs a rebase.

samwilson:
Rebased.

PR #170, opened 2019-05-13
Reorganise classes to PSR4

samwilson:
* Make sure each class is in its own file, and move all under the `App\` namespace.
* Autoload classes, and remove init files and all `require` statements.
* Move utility functions into a new `App\Util\Util` class.
* Move files into `src/`, `public/`, `resources/`, and `bin/` directories.


MaxSem:
* Why put the exception into the `Util` namespace, especially since another one is in a different place?
* Strict types are a very major change that can introduce lots of failures - is it necessary for this commit?

samwilson:
Good points, thanks. Have fixed both things.

@Tpt do you think this change is useful?

MaxSem:
Huh, this PR got old. Sam, if you rebase it I think it can be merged.

samwilson:
Rebased and ready to go I think.

samwilson:
Not ready to go. I'm not sure why epubcheck is failing all of a sudden. :-( I'll look more into this tomorrow...

samwilson:
Oh, epubcheck has been failing forever. Those errors are just not actual test failures.

It's passing now.

MaxSem:
Emm, I don't think I have +2 here, but as far as I'm concerned, it's +2 :)

samwilson:
Thanks!

Issue #168, opened 2019-05-07
Don't store usage logs in the temp directory


samwilson:
Currently, `temp/logs.sqlite` stores the book-generation statistics. This file is not temporary and should be kept somewhere in which it's less likely to be accidentally deleted.

samwilson:
Given that we have a symlink from `http/` to the log sqlite database file, why don't we just store the logs in `http/` to start with? There's nothing private about them.

At the moment we have `logs.sqlite` (116 MB) and `logs.sqlite.backup` (70 MB, I don't know the history of this file). Maybe one day if the log file gets too big we could move to MySQL, but for now it's probably fine (unless it's also bad to have it on NFS?).

Tpt:
Moving logs to http/ seems fine to me.

`logs.sqlite.backup` contains the historical logs from the wsexport.wmflabs.org VPS server.

Using MariaDB would be indeed better for NFS performances but we would have to provide a way to easily query the database (maybe by making it public and so queryable using Quarry?)

samwilson:
Okay, logs.sqlite is now in http directory. Test and prod sites both updated with the new config. Everything seems to be working fine.

Switching to MariaDB can happen in a separate ticket I think. I think you're right @tpt, a public tool DB that's queryable from Quarry would be great.

samwilson:
I've created #173 for the remaining DB stuff.

PR #171, opened 2019-05-15
Move logs.sqlite to the http directory

samwilson:
Add a new config variable 'logDatabase' that specifies the
full filesystem path to the logs.sqlite file in which to store
access statistics.

Also switch to using the new config.php file in stats.php so
that this new variable need only be set in one place.

Ref: #168

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ temp/*
 vendor/*
 composer.phar
 config.php
+*.sqlite
+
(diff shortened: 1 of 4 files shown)


samwilson:
Note that upgrading to this requires manually moving the existing log file. This means we'll have to stop the webservice and move the file before upgrading to this. It seems not too bad to do that, rather than add additional complexity here to make the application handle the upgrade itself.

Issue #153, opened 2019-03-22
Consider using RemexHTML


Tpt:
RemexHTML is a PHP HTML parser written by the WMF. We may consider using it instead of relying on the buggy DOM PHP implementation.

https://www.mediawiki.org/wiki/RemexHtml

We may also have a look at https://www.mediawiki.org/wiki/HtmlFormatter

samwilson:
I've moved this issue to https://phabricator.wikimedia.org/T259096

PR #172, opened 2019-05-15
Handle Wikimedia server errors nicely

MaxSem:
Bug: https://phabricator.wikimedia.org/T222932

### DIFF
--- a/utils/Api.php
+++ b/utils/Api.php
@@ -7,7 +7,6 @@
 
 use GuzzleHttp\Client;
 use GuzzleHttp\ClientInterface;
-use GuzzleHttp\Exception\RequestException;
 use GuzzleHttp\HandlerStack;
 use GuzzleHttp\Pool;
 use GuzzleHttp\Promise\PromiseInterface;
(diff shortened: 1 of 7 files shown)


samwilson:
With regard to using HtmlFormatter, do you want to comment on #153? Seems that maybe it will do everything we need for parsing MediaWiki HTML elsewhere in this tool as well, so maybe that can be resolved.

MaxSem:
I'm using HtmlFormatter only to read error HTML into DOM, the question in that ticket is much more global.

utils/utils.php
```diff
@@ -197,3 +200,43 @@ function formatException( Exception $ex ): string {
...
+	}
+
+	// Otherwise, try hhvm-fatal-error.php style
+	if ( !$text ) {
+		foreach ( $doc->getElementsByTagName( 'code' ) as $node ) {
+			$text = trim( $text . "\n" . $node->textContent );
+		}
+	}
+
+	return $text ? "Wikivoyage servers returned an error: $text" : null;
```
samwilson: Wikivoyage? :)

utils/Api.php
```diff
@@ -76,10 +75,10 @@ function ( ResponseInterface $response ) {
 					throw new HttpException( 'HTTP error ' . $response->getStatusCode(), $response->getStatusCode() );
 				}
 				return $response->getBody()->getContents();
-			},
+			}/*,
 			function ( RequestException $e ) {
```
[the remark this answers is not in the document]
MaxSem: Done.

utils/utils.php
MaxSem: You don't like Wikivoyage, eh?

PR #175, opened 2019-06-01
Rework cache operation

MaxSem:
Currently, wsexport attempts to clean after itself, e.g. in
Picture::__destruct(), however this doesn't protect us from
some types of fatal errors, PHP segfaults and programmer
mistakes. Since we want to keep the temporary files in local
/tmp to make I/O fast, we can't clean them up via a gronjob,
as the webservice can be run on different lighttpd VMs.

This patch introduces a new class FileCache and makes
everything use it:
* It maintains a per-tool directory under tempPath
  (which is now intended to point to /tmp on Toolforge).
* Cleans this directory of old files on 5% of requests.
* Unifies all places using temporary files/directories, making
  it use the new class. Util::buildTemporaryFileName() can't
  override this now (and thus it can't forget files in places
  FileCache doesn't clean).

Bug: https://phabricator.wikimedia.org/T166337

### DIFF
--- a/bin/cleanup.sh
+++ b/bin/cleanup.sh
@@ -1,8 +1,0 @@
-#!/usr/bin/env bash
-# Run periodically via cron to delete temporary files
-
-DIR=$(dirname $(dirname $0))
-
-find "${DIR}/temp" -name "ws-*" -mtime +1 -delete
-
-find "${DIR}/temp" -name "img-*" -mtime +1 -delete
--- a/src/BookProvider.php
+++ b/src/BookProvider.php
@@ -212,9 +212,9 @@ class BookProvider {
 	protected function getPicturesData( array $pictures ) {
 		$promises = array_map(
 			function ( Picture $picture ) {
-				global $wsexportConfig;
+				$cache = FileCache::singleton();
 
-				$picture->tempFile = tempnam( $wsexportConfig['tempPath'], 'pic-' );
+				$picture->tempFile = tempnam( $cache->getDirectory(), 'pic-' );
 				$options = [ 'sink' => $picture->tempFile ];
 				return $this->api->createAsyncRequest( $picture->url, $options );
 			},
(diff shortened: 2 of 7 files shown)


aezell:
@MaxSem If you rebase this with the conflicting changes, I'll take a look and review it.

utils/FileCache.php
```diff
@@ -0,0 +1,111 @@
...
+		// Use username in the path. The tool needs its own directory so that it knows which
+		// files to garbage collect. Can't just use the hardcoded name because
+		// wsexport and wsexport-test might end up on the same node, stepping on each other's toes.
+		$user = get_current_user();
+		// Remove leading "tools." present if running on labs
+		$userCleaned = preg_replace( '/^tools\./', '', $user );
+		if ( $user !== $userCleaned ) {
+			// We're on Toolforge, gotta compartmentalize
+			$dir = $path . '/' . $userCleaned;
+		}
```
Tpt: Is it important of doing it only on labs? We might want to do it everywhere to make code simpler.

utils/FileCache.php
```diff
@@ -0,0 +1,111 @@
...
+			$this->statFile( $fileName );
+		}
+	}
+
+	/**
+	 * Checks whether the given file is present locally and not stale.
+	 * Deletes the file if it's stale.
+	 * @param string $fileName
+	 */
+	protected function statFile( string $fileName ): void {
```
Tpt: There seem to me there is a typo in the method name. It should be something like `staleFile`.

utils/FileCache.php
MaxSem: Done.

utils/FileCache.php
MaxSem: No typo because it's intended to mean "run `stat` on file", but yeah - it's bad and I'll fix it.

MaxSem:
Thanks for the review! The files are already getting deleted in `Picture::__destruct()` this is just another measure in case some files remain due to a fatal crash or segfault, I'll elaborate this in commit message.

MaxSem:
This is ready for re-review.

Issue #173, opened 2019-05-15
Move access logs to Toolforge DB


samwilson:
The access logs are currently stored in a Sqlite database with the following structure:

```sql
CREATE TABLE `creation` (
	`lang` VARCHAR(40) NOT NULL,
	`title` VARCHAR(200) NOT NULL,
	`format` VARCHAR(10) NOT NULL,
	`time` DATETIME  NOT NULL
);
```

These should be moved to an equivalent MariaDB database so that they can be queried from Quarry, and be better for multiple writes from the web tool.

samwilson:
PR for this issue is #184.

samwilson:
This is done, and I'm deploying now.

samwilson:
* Test database: `s54020__wsexport_test_p`
* Prod database: `s52561__wsexport_p`

The Sqlite-MariaDB import process is now running.

samwilson:
Okay, both old logs.sqlite files have now been imported, for a total of 1,404,994 records (not all of which are imports; I turned the webservice back on before the import had finished, so there is some number of today's stats in that too).

All seems good, but let me know if you find anything amiss! I'll keep an eye on things of course.

PR #184, opened 2019-06-19
Move logs to MariaDB

samwilson:
* Add tests for logging class.
* Create import tool for Sqlite →  MariaDB.
* Documentation (note that bin/install.php needs to be run)

Refs #173
Bug: https://phabricator.wikimedia.org/T226070

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,18 @@ dist: trusty
 php:
   - '7.2'
 
+services:
+  - mysql
+
 before_install:
   - sudo apt-get update -q
   - sudo apt-get install calibre-bin default-jre -y
 
 install:
   - make install
+  - mysql -e "CREATE DATABASE wsexport_test;" -uroot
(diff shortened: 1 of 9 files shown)


samwilson:
Prior to this patch, because tests will now need their own config, I'd like to add #185. It'll simplify this patch.

Tpt:
Thank you! It would also be nice to import the log file `logs-wsexport.wmflabs.org-instance.sqlite` from the previous instance (it is stored in the root directory of the wsexport tool)

samwilson:
>  It would also be nice to import the log file logs-wsexport.wmflabs.org-instance.sqlite

Yes, I agree. My idea was (and sorry for not making this clear in this PR) to import whatever's pointed to by the `logDatabase` config value. So we could either change that to point to the other file, or rename the other file. The import script here is idempotent, so it should be fine even if there's some overlap or something goes wrong and we have to restart the import.

I'll bring this patch up to date.

PR #185, opened 2019-06-21
Universal bootstrap for all entrypoints

samwilson:
There are five entrypoints to this application: two web, two CLI,
and one test. The test one loads the others sometimes, and in
doing can have its config clobbered. This change moves all
config-loading to a single file, bootstrap.php, which can then
avoid double-loading.


samwilson:
Thanks @tpt :-)

PR #187, opened 2019-06-25
Fully-qualify URLs in About page, and remove toc

samwilson:
This fixes two things that epubcheck was complaining about:

1. "Found a reference to a resource that is not a spine item."
   with the ToC being referenced with `href="#toc"`, which isn't
   permitted. We can't link directly to the actual content-ToC
   because that's not in a standard place in Wikisource works.
2. "Referenced resource could not be found in the EPUB." with
   links in the About page that are not fully qualified. Fixed
   by adding the protocol and domain name.


Tpt:
Thanks!

PR #188, opened 2019-06-25
Update dependencies and remove an out of date phpcs exclusion

samwilson:
This updates Composer dependencies to their latest versions,
and removes the OneObjectStructurePerFile phpcs rule exclusion
which has been unnecessary since we moved to PSR4.


samwilson:
Thanks @mooeypoo :-)

PR #191, opened 2019-07-04
Increase exec timeout to 120s

MaxSem:
Bug: T227245


aezell:
And I see it's already merged. Classy.

PR #192, opened 2019-07-17
Sanitize images' mime types

samwilson:
We were using the full Content-Type header as the media-type
of epub manifest items, but this attribute is somewhat more
restricted in what it'll accept. So we discard everything after
the first semicolon.

https://phabricator.wikimedia.org/T227334

Bug: T227334


samwilson:
epubcheck was saying this about the full content-type string:

> Non-standard image resource of type image/svg+xml; charset=utf-8; profile='https://www.mediawiki.org/wiki/Specs/SVG/1.0.0' found.

Now it's happy.

PR #193, opened 2019-07-17
Actually check for Java

samwilson:
Check whether java is installed before running epubcheck tests.


samwilson:
Oops. Sorry, I introduced this bug with the process refactoring.

PR #197, opened 2019-08-20
Add recent popular books to front page

samwilson:
Add a list of books popular in the last 3 months to the left side
of the main page. Also add a breadcrub link back to the main page
from the stats page.

This patch indents a bunch of lines; it's easier to see the changes if you use Github's "ignore whitespace changes" option.

### DIFF
--- a/public/templates/book.php
+++ b/public/templates/book.php
@@ -1,81 +1,99 @@
 <?php
 
+use App\CreationLog;
 use App\FontProvider;
 use App\GeneratorSelector;
 use App\Util\Api;
 
 include 'header.php';
 ?>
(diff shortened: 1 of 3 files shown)


samwilson:
Thanks @tpt!

zdzis:
@samwilson @Tpt, maybe it would be better (for _marketing_ reasons) in the list of popular books to give a link to the page in Wikisource instead of the direct download link?
What do you think?
Z.

Tpt:
@zdzis Indeed, having both a link to Wikisource and a download links seems relevant to me.

samwilson:
Good idea! I'll update it.

samwilson:
See #203.

PR #202, opened 2019-09-05
Expand page title

samwilson:
Add a descriptive title to the page, for greater clarity and to
avoid the Scunthorpe problem.

### DIFF
--- a/public/templates/header.php
+++ b/public/templates/header.php
@@ -3,7 +3,7 @@
 <html lang="en">
 <head>
 	<meta charset="UTF-8"/>
-	<title>WSexport</title>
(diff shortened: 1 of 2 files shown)


MaxSem:
Can we avoid calling everything that's not MediaWiki a "tool"? That doesn't add any clarity to the name.

samwilson:
I guess it's just to reinforce that it's on Toolforge? But you're right, it doesn't really need it.

The page's H1 is "Export tool of Wikisource books in many file formats."; I was just going off that.

samwilson:
I've removed 'tool' and added a page subtitle paragraph.

PR #211, opened 2019-11-06
Makes sure that templatestyles are properly used

### LINKED ISSUE #210: Adds templatestyle support
Wsexport should keep templatestyle style sheet in the output.

Tpt:
* Moves body <style> to the <head> node and make sure it's used on all splits of a file.
* Makes sure that <body> attributes are preserved during split

Closes #210


Tpt:
Let's get things out

PR #212, opened 2019-11-21
Move query that display stats in home page to statistics page

cimurah:
Improve performance by moving expensive DB query that diplays some stats
to the statistics page (stats.php).

Bug: T236391
Bug: T237029

### DIFF
--- a/public/templates/book.php
+++ b/public/templates/book.php
@@ -1,106 +1,84 @@
 <?php
 
-use App\CreationLog;
 use App\FontProvider;
 use App\GeneratorSelector;
 use App\Util\Api;
 
 include 'header.php';
 ?>
-<div class="row">
-	<form method="get" action="book.php" role="form" class="form-horizontal col-md-9">
-		<fieldset>
-			<legend>Export a file</legend>
-			<div class="form-group">
-				<label for="lang" class="col-lg-2 control-label">Language code</label>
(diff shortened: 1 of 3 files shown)


public/templates/book.php
```diff
@@ -1,31 +1,34 @@
 <?php
 
-use App\CreationLog;
 use App\FontProvider;
 use App\GeneratorSelector;
 use App\Util\Api;
 
 include 'header.php';
 ?>
+<ol class="breadcrumb">
```
samwilson: The 'breadcrumb' is used to show the current position in a hierarchy, not a navigation menu. I think it could just be left out here, because there's already a link to the statisics in the footer.

public/templates/book.php
```diff
@@ -1,31 +1,34 @@
...
+	<li class="active">Home</li>
+	<li><a href="stat.php">Statistics</a></li>
+</ol>
 <div class="row">
```
samwilson: Now that there's only one column, this whole row element could be removed (and the col specification below). Not that it really matters.

samwilson:
@cimurah not sure if you saw my comments above.

public/templates/book.php
cimurah: Breakcrumb removed, thank you for the clarification :)

public/templates/book.php
cimurah: Done.

PR #215, opened 2020-02-11
Upgrades dependencies

Tpt:
We are now using PHP 7.3 on production servers. We could upgrade to it.
This allows to use the latest version of PHPUnit.


composer.json
```diff
@@ -18,7 +18,7 @@
 		}
 	},
 	"require": {
-		"php": ">=7.2",
+		"php": ">=7.3",
```
samwilson: None of these changes actually *require* PHP 7.3 do they?

I wonder if it might not be better to keep the lower version so that it's easier for developers to get wsexport running locally (7.2 or earlier is still standard in lots of distros).

composer.json
```diff
@@ -27,16 +27,16 @@
...
 		"wikimedia/html-formatter": "^1.0"
 	},
 	"require-dev": {
 		"ext-xdebug": "*",
-		"mediawiki/mediawiki-codesniffer": "~28.0",
-		"mediawiki/minus-x": "^0.3",
-		"phpunit/phpunit": "~8.0"
+		"mediawiki/mediawiki-codesniffer": "^29.0",
+		"mediawiki/minus-x": "^1.0",
+		"phpunit/phpunit": "^9.0"
```
samwilson: This could be `^8.0|^9.0`, to maintain PHP 7.2 compatibility but still permit PHP 7.3.

Tpt: Indeed. Done

composer.json
Tpt: Done.

Tpt:
I modified the change to still keep PHP 7.2 compatibility. The `composer.lock` is the one for PHP 7.3 because it is the version used on production server.

Tpt:
`composer install` does not like to have different dependencies for different PHP versions if we check-in the `composer.lock`. I modified the change to keep using PHP 8

PR #219, opened 2020-02-15
Use pdf-page-margin-* instead of margin-* for PDF generation

samwilson:
The PDF generation commands for ebook-convert have their own
margin-setting CLI options.

Bug: https://phabricator.wikimedia.org/T244825


samwilson:
The [docs](https://manual.calibre-ebook.com/generated/en/ebook-convert.html) say "Page oriented formats such as PDF and DOCX have their own margin settings that take precedence."

PR #221, opened 2020-02-20
Uses just a plain image as cover

### LINKED ISSUE #220: Cover image displayed at the end of the PDF with Calibre 3.46 and Calibre 4
The ePub cover images seems to be displayed at the end of the converted PDF with the recent versions of Calibre. It seems to be caused by the fact that the cover is included as an XHTML item and  in the `spine` in the `content.opf` file but is tagged with `linear="no"`.

A simple solution might be to completely drop the `cover.xhtml` file and only keep the cover image as an image.

Tpt:
Closes #220


Tpt:
@samwilson May you have a look at this change?

Tpt:
Let's merge this and see how it goes on the test server.

PR #228, opened 2020-05-10
Drops FileCleaner

Tpt:
Does not work for fixing the jpg problem and the SVG fix is not useful now that we dropped ePub 2 support


samwilson:
Looks good to me!

PR #230, opened 2020-06-12
Switch to using system fonts

samwilson:
Rather than including our own fonts, instead include fonts from
the host system.

https://phabricator.wikimedia.org/T261479


samwilson:
> Code looks OK to me, though I admit I don't really understand how the system works. I was going to test this on https://wsexport-test.wmflabs.org/, but first things first -- I can't repro the original issue. On my end, all characters appear correctly when exporting [this book](https://kn.wikisource.org/wiki/%E0%B2%B5%E0%B2%BF%E0%B2%AE%E0%B3%8B%E0%B2%9A%E0%B2%A8%E0%B3%86) (linked to the description for [T254918](https://phabricator.wikimedia.org/T254918)) on production. I do not see any boxes. Are there other examples I could try?

You might have fonts installed that support Kannada. I think the issue is most likely to come up when a book is exported on a system with the right fonts (and so everything looks fine) is transferred to a device (e.g. an ereader) that doesn't have the right fonts. With this fix, the user will be given the option to embed the correct font, but they'll still have to actually do that to avoid the issue.

> 
> I am still going to +1 because I trust you :)

hehe thanks. I'll make sure not to put it into production without good testing on the test site!

samwilson:
New branch for this: #248

PR #231, opened 2020-06-12
Add php-imagick to Travis CI

samwilson:


samwilson:
Hmm seems to still be erroring: https://travis-ci.org/github/wsexport/tool/jobs/697520085#L981

PR #234, opened 2020-06-29
Add docker dev environment

dayllanmaza:
Everything seems to work except for integration tests. I might need some help with that.

Apache config can be changed to remove "/public/" from the path

Not sure what else I could be missing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,8 @@ RUN apt-get update -q && apt-get install -y \
         libzip-dev \
         unzip \
         wget \
+        xdg-utils \
+        xz-utils \
       && rm -rf /var/lib/apt/lists/* \
       && pecl install xdebug zip \
       && docker-php-ext-enable xdebug zip \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,26 @@
+FROM php:7.2-apache-stretch
+
+WORKDIR /var/www/html
+
+COPY ./docker/000-default.conf /etc/apache2/sites-available/
+COPY ./docker/ports.conf /etc/apache2/
+
+EXPOSE 8080
+
+# work around https://github.com/docker-library/openjdk/blob/0584b2804ed12dca7c5e264b5fc55fc07a3ac148/8-jre/slim/Dockerfile#L51-L54
+RUN mkdir -p /usr/share/man/man1
+
+RUN apt-get update -q && apt-get install -y \
+        epubcheck \
+        libzip-dev \
+        unzip \
+        wget \
+      && rm -rf /var/lib/apt/lists/* \
+      && pecl install xdebug zip \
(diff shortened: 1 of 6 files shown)


samwilson (changes requested):
This looks good!

It sounds like the test error is a missing dependency:

> ImportError: libGL.so.1: cannot open shared object file: No such file or directory

Don't worry about the all the phpunit *warnings*, they're always there (we've got issues open for quite a few of them now I think). The errors should be able to be fixed though.

Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+EXPOSE 8080
+
+# work around https://github.com/docker-library/openjdk/blob/0584b2804ed12dca7c5e264b5fc55fc07a3ac148/8-jre/slim/Dockerfile#L51-L54
+RUN mkdir -p /usr/share/man/man1
+
+RUN apt-get update -q && apt-get install -y \
+        epubcheck \
+        libzip-dev \
+        unzip \
+        wget \
```
samwilson, line 17: Looks like this also needs `libgl1-mesa-glx` (in my testing, that seems to get rid of the errors; there are still 38 warnings).

samwilson: Actually, sorry, scratch that, it should be these two (as listed in the Calibre install docs):

```
xdg-utils \
xz-utils \
```

docker/000-default.conf
```diff
@@ -0,0 +1,31 @@
...
+	# the server uses to identify itself. This is used when creating
+	# redirection URLs. In the context of virtual hosts, the ServerName
+	# specifies what hostname must appear in the request's Host: header to
+	# match this virtual host. For the default virtual host (this file) this
+	# value is not decisive as it is used as a last resort host regardless.
+	# However, you must set it for any further virtual host explicitly.
+	#ServerName www.example.com
+
+	ServerAdmin webmaster@localhost
+	DocumentRoot /var/www/html
```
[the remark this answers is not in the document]
dayllanmaza: Thank you. Fixed

Dockerfile
dayllanmaza: I've added both. Thanks

README.md
```diff
@@ -63,6 +63,62 @@ $ ./vendor/bin/phpunit --exclude-group integration
...
+'dbDsn' => 'mysql:host=database;dbname=wsexport;charset=utf8',
+'dbUser' => 'root',
+'dbPass' => '',
+  ```
+
+```bash
+docker-compose exec wsexport php ./bin/install.php
+```
+
+Wikisource export should be up at http://localhost:8888/public/book.php
```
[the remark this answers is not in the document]
dayllanmaza: Done

MusikAnimal (approved):
LGTM!

PR #242, opened 2020-07-16
Switch to i18n-friendly long date format for cover

samwilson:
Bug: T256986

### DIFF
--- a/src/Generator/EpubGenerator.php
+++ b/src/Generator/EpubGenerator.php
@@ -7,6 +7,7 @@ use App\Cleaner\BookCleanerEpub;
 use App\FontProvider;
 use App\Util\Util;
 use Exception;
+use IntlDateFormatter;
 use ZipArchive;
 
 /**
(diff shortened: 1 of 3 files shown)


src/Generator/EpubGenerator.php
```diff
@@ -50,7 +51,7 @@ public function create( Book $book ) {
 		$oldBookTitle = $book->title;
 		$css = $this->getCss( $book );
 		$this->i18n = Util::getI18n( $book->lang );
-		setlocale( LC_TIME, $book->lang . '_' . strtoupper( $book->lang ) . '.utf8' );
+		setlocale( LC_TIME, $book->getLocale() . '.utf8' );
```
Tpt, line 54: We could probably remove this if we do not have anything else relying on the current locale.

src/Generator/EpubGenerator.php
```diff
@@ -339,6 +340,7 @@ private function getXhtmlTitle( Book $book ) {
 			$footerElements[] = $book->year;
 		}
 
+		$formatter = new IntlDateFormatter( $book->getLocale(), IntlDateFormatter::LONG, IntlDateFormatter::NONE );
```
Tpt, line 343: It seems to work properly with just `$book->lang` instead of `$book->getLocale()`. It would allow ICU to choose the most common locale for each language and not existing locales like `en_EN`.

src/Generator/EpubGenerator.php
samwilson: Good point, done.

src/Generator/EpubGenerator.php
samwilson: Oh yes, that works great! Thanks.

PR #244, opened 2020-10-02
Switch to Symfony framework

samwilson:
This patch migrates the entire application to Symfony, and tries to modify the 
smallest amount of code possible to make this work.

Some notes:

* Tests in CI were already not passing, so the failing tests are now excluded 
(via the `exclude-from-ci` test group). These are all to do with epub standards 
conformance, and will be tackled later.
* The CLI entry point bin/book.php has been deleted, which might impact some 
people's scripts.
* A bunch of bits of old code have been moved to Controllers, Commands, or 
templates, without change. These are sometimes showing as changes in the Git 
diff, but actually still need to be brought up to date.
* Fully setting up Doctrine isn't done yet. Although the single DB table we 
need could be done via a migration now, it's not yet able to be used with 
Doctrine ORM because it doesn't contain a simple primary key, so it seems 
easier to leave this for a subsequent patch.
* There is duplication in tests for configuring the database, but this seems 
acceptable (rather than refactoring it to somewhere common) because it'll all 
be removed after the above.
* Docker Compose operation is complicated now by the need to split environment
variables between `.env` and `.env.local`.
* HTTP status text is not working correctly (but the response code is).

Bug: T263577


README.md
```diff
@@ -24,7 +24,7 @@ Installation
 
        composer install --no-dev
 
-   This will create a `config.php` file that you can edit.
+   Then create a `.env.local` file.
```
dayllanmaza, line 27: Is there a reason why docker and symfony can't share the same .env file?

dayllanmaza (commented):
Looks great. I'm still going thru it while I look into symfony a bit more. I'm leaving a couple comments in the meantime

composer.json
```diff
@@ -18,42 +18,87 @@
...
+		"symfony/console": "5.1.*",
+		"symfony/dotenv": "5.1.*",
+		"symfony/flex": "^1.3.1",
+		"symfony/framework-bundle": "5.1.*",
+		"symfony/process": "5.1.*",
+		"symfony/yaml": "5.1.*",
+		"twig/extra-bundle": "^2.12 || ^3.0",
+		"twig/twig": "^2.12 || ^3.0",
+		"wikimedia/html-formatter": "^1.0",
+		"wikimedia/toolforge-bundle": "^1.0"
```
dayllanmaza, line 45: wikimedia/toolforge-bundle 1.0.0 requires ext-intl

You need to add 
```
libicu-dev \
...
docker-php-ext-install intl \
```
to the Dockerfile to satisfy the requirements (I think)

README.md
dayllanmaza: I see. I do not want to stop this PR because of docker unless someone other than me is using it. @cimurah @MusikAnimal what do you think? Should we fix docker in a separate patch? Do you want to support it at all?

dayllanmaza (approved):
We've agreed to work on the docker bits in a separate patch.

PR #245, opened 2020-10-04
Do not include cover images if the "include images" option is disabled

Tpt:
Task: https://phabricator.wikimedia.org/T264524


samwilson (approved):
Looks good!

PR #248, opened 2020-10-14
Switch to using system fonts

samwilson:
Rather than including our own fonts, instead include fonts from
the host system.

Get the font info with the fc-list command, and cache it for a
week.

Turn the FontProvider into a Symfony service class and inject it
everwhere required. This makes it easier to test, and will make
more sense as we progressively convert more code to the Symfony
architecture.

This patch also makes it possible to provide *any* available
font name via the URL, e.g. `?fonts=WikisourceMono`, even if
the name isn't listed in the dropdown in the export form.

Bug: T261479


src/FontProvider.php
```diff
@@ -13,96 +18,118 @@
...
 
 	/**
-	 * return CSS
+	 * Get CSS for the given font.
+	 * @param string $id The font ID as defined in this class.
 	 * @return string
 	 */
-	public static function getCss( $id, $basePath ) {
-		if ( !isset( self::$data[$id] ) ) {
+	public function getCss( string $id ): string {
```
dayllanmaza, line 113: **Question**: This method used to assign all font families even if not defined/included. Now it seems to only do that for the one that is selected by the user. I'm guessing this new way is intentional and a bit of a cleanup?
I also see body defined twice in the final css. Not sure if that's a problem
```
body { font-family: "FreeSerif" }
body {
	line-height: 1.2;
}
```

dayllanmaza (changes requested):
Looks good so far. I have a couple minor comments. I'm still to go thru it all.
Works great except for the default font on non-latin langs

dayllanmaza:
It would be great if this can be rebased on top of `dev` so we can use the new docker setup to test it. At least the following should be added to the Dockerfile
```
        fontconfig \
        fonts-freefont-ttf \
```

docker/Dockerfile
```diff
@@ -24,6 +24,10 @@ RUN apt-get update -q && apt-get install -y \
         xdg-utils \
         xz-utils \
         libicu-dev \
+        fonts-freefont-ttf \
```
dayllanmaza, line 27: missing fontconfig to add fc-list to the container

src/FontProvider.php
```diff
@@ -13,96 +18,181 @@
...
-		'mukta-vaani' => [
-			'name' => 'Mukta', 'label' => 'Mukta Vaani (Gujarati)', 'css_name' => 'wse_MuktaVaani', 'otf' => [
-				'R' => 'MuktaVaani-Regular.otf', 'RB' => 'MuktaVaani-Bold.otf'
-			]
-		]
-	];
+	/** @var CacheInterface */
+	private $cache;
+
+	/** @var string[][] */
```
dayllanmaza, line 24: Nit: Technically this can be null.

src/Controller/ExportController.php
```diff
@@ -129,11 +135,14 @@ private function export( Request $request, CreationLog $creationLog, LoggerInter
 	 * @param string $lang A language code.
 	 * @return string
 	 */
-	private function getFont( Request $request, $lang ): ?string {
+	private function getFont( Request $request, $lang, FontProvider $fontProvider ): ?string {
 		// Default font for non-latin languages.
-		$font = $request->get( 'fonts', '' );
+		$font = $fontProvider->resolveName( $request->get( 'fonts', '' ) );
```
dayllanmaza, line 140: Maybe I'm missing something but `$fontProvider->getOne()` calls `resolveName()` internally so this call is unnecessary. If that's the case, `FontProvider::resolveName()` can be set as private. Does that makes sense?

dayllanmaza (changes requested):
Works great. I have a couple tiny comments and it should be good to go. Nice job

src/FontProvider.php
samwilson: Done.

src/FontProvider.php
dayllanmaza: I think it doesn't really matter. Mentioned it in case it was an easy fix

src/Controller/ExportController.php
dayllanmaza: Good idea about calling it inside `getCss()`. I'm afraid I still don't understand why it needs to be called in the controller and it looks to me that it should be an implementation detail of `FontProvider` but I might be missing something. LGTM otherwise

dayllanmaza (approved):

PR #254, opened 2020-10-20
Add logging, including emailing fatals in prod

samwilson:
Add Monolog and Symfony Mailer, the latter for emailing fatal
errors from production to the tool's maintainers' email address.


MusikAnimal (approved):
Thank you for this! Works great on my local. I tried testing it out on staging but could not confirm an email was sent. I did this by changing `DateTime` to `Foo` in StatisticsController.php line 24. I did correctly see the application-level 500 error page and exceptions logged to prod.log, though. I thought maybe I didn't get an email because I'm not a direct maintainer of tools.wsexport (emails don't follow group accounts like tools.community-tech-tools, right?), but I didn't get an email after switching the to/from to `tools.copypatrol` either. Maybe I was doing something wrong; I was sure to clear the cache for prod, set ownership of the files back to www-data, etc. (just like the upgrade.sh script). I've switched back to the `dev` branch but the `monolog` branch is still on wsexport-dev01 if you want to test it. If you're confident let's go ahead and merge this PR.

samwilson:
Hmm, I'm not sure! It's giving this error:

> Uncaught Symfony\\Component\\Mailer\\Exception\\TransportException: Connection could not be established with host "ssl://mail.tools.wmflabs.org:465": stream_socket_client(): unable to connect to ssl://mail.tools.wmflabs.org:465 (Connection timed out)

According to [the docs](https://wikitech.wikimedia.org/wiki/Help:Toolforge/Email#Mail_from_Tools), it should work.

I changed to port 587 and got the same.  Port 25 doesn't give any error, but doesn't send any mail. :-(

Sending mail from the command line with `/usr/sbin/exim -odf -i` works fine, so I tried changing to `MAILER_DSN=sendmail://default` but got this:

> Uncaught Symfony\\Component\\Mailer\\Exception\\TransportException: Expected response code "250/251/252" but got code "550", with message "550 Administrative prohibition".

Oh!! I get it now...

It's like you say: we're not direct maintainers of the tool and don't get mail to tools.wsexport@tools.wmflabs.org!! :-( 

> but I didn't get an email after switching the to/from to tools.copypatrol either

Did you clear the cache right after that? That's the trouble I was having, it looks like.

All up, this does seem to work! The only change was the port number in the DSN; I'll update `.env` to have that.

samwilson:
To fix the recipients issue, rather than adding us all as maintainers it seems neater to just add the commtech tools email address. It should work by adding a JSON-encoded array as the env var:

    APP_LOG_RECIPIENTS='["tools.wsexport@tools.wmflabs.org", "tools.community-tech-tools@tools.wmflabs.org"]'

and reading it in the config with the `json` processor:

    to_email: '%env(json:APP_LOG_RECIPIENTS)%'

…but there's [a bug](https://github.com/symfony/monolog-bundle/issues/237) in the Monolog bundle that prevents this from working.

Instead, I've added two env vars, and in development we can just set them both to the same address. Is this too messy? See what you think.

MusikAnimal:
Sorry for taking so long to get back to you. With the latest push I'm now getting a `ProcessFailedException` when loading any/every page:

```
The command "fc-list --format "%{file};%{family};%{style};%{lang}\n"" failed.

Exit Code: 127(Command not found)

Working directory: /var/www/html/public

Output:
================


Error Output:
================
sh: 1: fc-list: not found
```

I don't know if it's an issue with Docker. Does this happen for you, too? I'm currently unable to test outside of Docker because my system is PHP 7.4.

MusikAnimal:
Never mind! After rebuilding the container all is working fine :)

> I've added two env vars, and in development we can just set them both to the same address. Is this too messy? See what you think.

I don't think it's that bad, considering the limitations we're faced with (we want tools.community-tech-tools to be emailed; the Monolog bug you mentioned). I saw the email too so I'll take that as confirmation it worked.

Merging!

PR #255, opened 2020-10-20
Remove custom error handling in favour of Symfony's

samwilson:
Remove the custom HttpException class and rearrange the error
handling in the ExportController. This makes it a bit simpler but
keeps the same responses as formerly.


src/Controller/ExportController.php
```diff
@@ -137,4 +110,33 @@ private function getFont( Request $request, $lang ): ?string {
...
+	 */
+	public function error( Throwable $exception, Api $api ) {
+		$message = $exception->getMessage();
+		if ( $exception instanceof RequestException ) {
+			$exceptionResponse = $exception->getResponse();
+			if ( $exceptionResponse ) {
+				$message = Util::extractErrorMessage( $exceptionResponse, $exception->getRequest() );
+			}
+		}
+		return $this->render( 'export.html.twig', [
```
MusikAnimal, line 129: Perhaps we could check the env and only do this in production? For 500s especially I think it's ideal that we can see the stack trace without having to check the logs. I also quite like Symfony's exception interface where it conveniently highlights the problematic lines of code.

samwilson: You don't have to check the logs, you can just click the debug toolbar and get it in a readable display. But I know what you mean! How about we just catch Http and Request exceptions? Those are more likely to be ones that the user can recover from (e.g. wrong page title or wiki lang code).

samwilson: I've updated it to just check Http and Request (the traces of which are still visible via the toolbar).

MusikAnimal (approved):
Looks/works great! I forgot about seeing the stack trace via the debug toolbar :) HttpException and RequestException should cover the recoverable stuff, so this works well I think.

PR #258, opened 2020-10-23
Make BookCreator injectable

samwilson:
Refactor the BookCreator class so that it can be injected as a
service. This is the first step towards making it easier to use
common Symfony features such as caching in various places in the
codebase.

Bug: T265660


samwilson:
No longer required.

PR #259, opened 2020-10-23
Changing trigger events on  ci to avoid duplicated checks

dayllanmaza:


.github/workflows/ci.yml
```diff
@@ -1,7 +1,12 @@
 name: CI
 
-on: [push, pull_request]
-
+on:
+  push:
+    branches:
+      - main
```
samwilson, line 6: And `dev`, for the time being?

dayllanmaza: right, yes

PR #262, opened 2020-10-27
Refactor Api::$lang

samwilson:
Move the method dealing with getting the lang from the request
into a controller, and add new accessor method to Api class. This
is a step towards making the Api class into a service.

Bug: T265660


cimurah (approved):
Looks good to me!

PR #263, opened 2020-10-27
Add Git status check to avoid changed files

samwilson:


samwilson:
I couldn't reproduce the bug I was getting earlier.

PR #264, opened 2020-10-27
Update Composer dependencies

samwilson:


samwilson:
We don't really need to do this now, I was just looking into what's changing the `symfony.lock` file on the test site. Might as well update though.

PR #266, opened 2020-10-27
Add caching to the API

samwilson:
Add caching to API calls and other URL-fetching, and an HTTP
cache as well. Also add a means to bypass this for web and CLI
export requests.

Bug: T265660

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,6 +33,7 @@
 		"doctrine/annotations": "^1.10",
 		"doctrine/doctrine-bundle": "^2.1",
 		"guzzlehttp/guzzle": "^6.0",
+		"kevinrob/guzzle-cache-middleware": "^3.3",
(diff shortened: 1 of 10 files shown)


samwilson:
This is failing because of the instantiation of `Api` in Util. #272 will fix this.

Tpt (approved):
Looks great to me! With this we might want in an other PR to clean up a bit the data handled by the `Refresh` class that might not be needed anymore now the the HTTP requests are cached. If we keep it, we need to make sure that the "refresh" call are also not using the HTTP cache.

samwilson:
Yes, that sounds like a good idea. We'll probably want to have different cache durations for different things (e.g. namespaces don't change often), but it'd be good to simplify things. Maybe we want have have separate `nocache` and `refreshcache` operations.

PR #270, opened 2020-10-28
Update font-handling for better selection, weights, etc.

samwilson:
The existing font family name handling wasn't taking into account
the fact that fc-list reports these with alternate names appearing
after a comma. This removes this trailing info, and also adds more
thorough weight and slant handling.

The dropdown list is also updated to include all available fonts,
because it sounds like at some point we'll update this to be more
responsive and useful but for now this makes it much easier to
test and see what is available.

Bug: T261479


cimurah:
This looks good to me and works in my testing.  I think it's ready to merge. Has anyone else taken a look? Please note that you might need to clear the cache by running `rm -r var/cache`; otherwise, you might get an error when selecting one of the new fonts.

PR #271, opened 2020-10-30
Add log email subject env var 

samwilson:
Add a new environment variable with which to change the subject
prefix of the emails sent when an error is logged.

Bug: T266849

### DIFF
--- a/config/packages/prod/monolog.yaml
+++ b/config/packages/prod/monolog.yaml
@@ -28,6 +28,6 @@ monolog:
             to_email:
                 - '%env(APP_LOG_RECIPIENT_1)%'
                 - '%env(APP_LOG_RECIPIENT_2)%'
-            subject: '[tools.wsexport] Fatal error: %%message%%'
+            subject: '%env(APP_LOG_SUBJECT)% %%message%%'
             formatter: monolog.formatter.html
             content_type: text/html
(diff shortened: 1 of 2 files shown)


config/packages/prod/monolog.yaml
```diff
@@ -28,6 +28,6 @@ monolog:
             to_email:
                 - '%env(APP_LOG_RECIPIENT_1)%'
                 - '%env(APP_LOG_RECIPIENT_2)%'
-            subject: '[tools.wsexport] Fatal error: %%message%%'
+            subject: '[%env(APP_TITLE)%] %%message%%'
```
dayllanmaza, line 31: This is fine; I have a question tho. I thought that the values that go into `.env` are values that you are not suppose to add to the repo (secrets, api keys, etc). Wouldn't it be more straight forward to just add the string to the yaml per environment? I'm new to symfony so this is the kind of questions that pops into my head :smile:

samwilson: That totally makes sense, but I'm not quite sure how to do it correctly because the test and prod sites are both running with APP_ENV as `prod`.

dayllanmaza: That's a good point. I don't know either. My guess is that whatever builds the staging env should copy prod configs and make the necessary changes but we don't have that kind of pipeline setup.

dayllanmaza: We talked for a bit about this at standup and we thought that it might be a good idea to have a separate env variable for the email subject instead of the app title so it can be more customizable (if case we need/want to). What do you think?

samwilson: Actually, I was just thinking about this last night: how about we add a new environment called `stage`? Then we can do what you first suggested and just have different configs for the different environments. I've updated the patch.

samwilson:
@dayllanmaza I've updated this to what we talked about.

PR #272, opened 2020-10-30
Remove instantiations of Api class

samwilson:
Remove the last instantiations of the Api class, in preparation
for being able to inject it everywhere (and inject it along with
other services like a cache and HTTP client; that'll come in a
subsequent patch).

Bug: T265660


tests/Book/RefreshTest.php
```diff
@@ -9,39 +9,48 @@
...
-class RefreshTest extends TestCase {
+class RefreshTest extends KernelTestCase {
+
+	/** @var Api */
+	private $api;
+
+	public function setUp(): void {
+		parent::setUp();
+		self::bootKernel();
+		$this->api = self::$kernel->getContainer()->get( Api::class );
```
dayllanmaza, line 25: With the following you can access private services in tests so you don't need to make them public
```suggestion
		$this->api = self::$container->get( Api::class );
```

config/services.yaml
```diff
@@ -32,6 +32,7 @@ services:
     App\Util\Api:
         calls:
             - setLogger: [ '@logger' ]
+        public: true
```
dayllanmaza, line 35: This is not necessary, Symfony tests have access to a container that includes both public and private services. See `RefreshTest.php` comment

tests/Book/BookProviderTest.php
```diff
@@ -64,7 +64,17 @@ public function testGetCreditsUsesCorrectToolServerURL() {
 	}
 
 	private function apiWithResponse( $status, $header, $body ) {
-		$responses = [ new Response( $status, $header, $body ) ];
+		// Set up all responses that are used by the above tests. Most don't need to return any data.
+		$responses = [
+			new Response( $status, $header, $body ),
```
dayllanmaza, line 69: I'm a bit confused as to why you need all these Response objects here. May I ask why is this necessary and just having the first line is not enough for the current tests?

src/Util/Util.php
```diff
@@ -74,20 +74,19 @@ public static function getXhtmlFromContent( $lang, $content, $title = ' ' ) {
 		return $html . '><head><meta content="application/xhtml+xml;charset=UTF-8" http-equiv="default-style" /><link type="text/css" rel="stylesheet" href="main.css" /><title>' . $title . '</title></head><body>' . $content . '</body></html>';
 	}
 
-	public static function getTempFile( $lang, $name ) {
+	public static function getTempFile( Api $api, $lang, $name ) {
```
dayllanmaza, line 77: will $lang ever be different than the language already set in the api instance? Looks to me that $lang was needed here 'cause the object was being constructed inside the method

dayllanmaza (commented):
I'm still wrapping my head around this but here are some things I found along with some questions if you don't mind

config/services.yaml
```diff
@@ -32,6 +32,7 @@ services:
     App\Util\Api:
         calls:
             - setLogger: [ '@logger' ]
```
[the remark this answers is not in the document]
samwilson: Good point. I've removed it, and am instead injecting a logger.

tests/Book/BookProviderTest.php
samwilson: umm... no idea! hehe

I've removed them. Everything seems fine. :-)

tests/Book/RefreshTest.php
samwilson: thanks! I figured I was missing something clever. :) Done.

tests/Book/BookProviderTest.php
samwilson: Oh, well: https://github.com/wsexport/tool/pull/272/checks?check_run_id=1422501542#step:7:189

I guess I tried the same thing last week!

src/Util/Util.php
cimurah: If you pass `beta` to Api, it sets the language as the user's default language but the temp files are saved under `beta` in the cache dir. I had the same question as @dayllanmaza but since I was testing using `beta`, I realized that this is a case where the Api lang is different than the $lang value.

tests/Book/RefreshTest.php
samwilson: Actually I'm not sure this makes any difference. I'm still getting "Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: The "App\Util\Api" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection instead."

src/Util/Util.php
dayllanmaza: @samwilson follow up patch is better yeah (if we need to change anything)
@cimurah I'm not sure I understand, can you elaborate more?

cimurah: There's some values that can be passed as the `$lang` like: `www`, `wl`, `wikilivres`, and `beta`. For instance, when we use the `beta` value, the Api.php makes the book request to `en.wikisource.beta.wmflabs.org` then it sets the `$lang = ''` so the temp files are grabbed from the `cached` root directory.

After analyzing this a bit further, it might be fine to use the `$lang` in the Api instance, but we should do it in a follow up since the logic around the cache is a bit convoluted.

dayllanmaza: oh I see, thank you for explaining. A follow up patch sounds good.

tests/Book/BookProviderTest.php
samwilson: Sorry, I'm not explaining it well! :) It's because in a clean install we have to run the Refresh process, and that takes another few requests. We don't care what the return, with the exception of the namespaces one. I've added a comment.

To replicate the failure, try deleting the cache files in your `var/` and running this test.

tests/Book/GeneratorSelectorTest.php
```diff
@@ -7,82 +7,92 @@
...
+	private $api;
+
+	/** @var GeneratorSelector */
+	private $generatorSelector;
+
 	public function setUp(): void {
 		parent::setUp();
 		$this->fontProvider = new FontProvider( new ArrayAdapter() );
+		self::bootKernel();
+		$this->api = self::$kernel->getContainer()->get( Api::class );
```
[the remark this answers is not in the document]
samwilson: Fixed.

samwilson:
Sorry for being slow to get back to this. It should be better now.

dayllanmaza:
> Sorry for being slow to get back to this. It should be better now.

it's fine. Looks good to me now

PR #274, opened 2020-10-30
Make Api::$domainName private

cimurah:
Make $domainName variable in Api class private and add getDomainName
method to make its value accessible.


cimurah:
Sorry @samwilson, I've created a new pull request because the old one was pointing to my [fork](https://github.com/cimurah/tool/tree/make-domainName-private).
I added your suggestion to this commit. 
> Needs /** @var string */ docblock.

PR #280, opened 2020-11-09
Fix duplicated font when family name is not consistent

dayllanmaza:


samwilson (approved):
I'm not completely certain that sorting alphabetically will work in 100% of cases (but I can't find any documentation about the ordering of the family names). Anyway, this fixes the issue at hand!

PR #281, opened 2020-11-18
Add symfony/stopwatch as a dev requirement

dayllanmaza:
Needed to see timing events on symfony profiler


samwilson:
Talking of time, what do you think of #275?

PR #283, opened 2020-11-27
Add PHP 8.0 to CI

samwilson:


samwilson:
We can revisit this later when upstream has updated more.

PR #286, opened 2020-12-03
Validate URL 'format' parameter

samwilson:
Avoid blank format param in ExportController and use a 404
exception instead of 500 for invalid format names.

Bug: T269343

### DIFF
--- a/src/Controller/ExportController.php
+++ b/src/Controller/ExportController.php
@@ -19,6 +19,7 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\ResponseHeaderBag;
 use Symfony\Component\HttpKernel\Exception\HttpException;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 // phpcs:ignore
 use Symfony\Component\Routing\Annotation\Route;
 use Throwable;
@@ -64,14 +65,13 @@ class ExportController extends AbstractController {
 		}
 
 		$title = $request->get( 'page' );
-		$format = $request->get( 'format', 'epub' );
 		$font = $this->getFont( $request, $api->getLang(), $fontProvider );
 		$images = (bool)$request->get( 'images', true );
 		return $this->render( 'export.html.twig', [
 			'fonts' => $fontProvider->getAll(),
 			'font' => $font,
 			'formats' => GeneratorSelector::$formats,
-			'format' => $format,
+			'format' => $this->getFormat( $request ),
 			'title' => $title,
 			'lang' => $api->getLang(),
 			'images' => $images,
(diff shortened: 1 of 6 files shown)


src/GeneratorSelector.php
```diff
@@ -46,7 +46,9 @@ public function getGenerator( $format ) {
 		} elseif ( $format === 'atom' ) {
 			return new AtomGenerator();
 		} else {
-			throw new WSExportInvalidArgumentException( "The file format '$format' is unknown." );
+			$msgFormat = '"%s" is not a valid format. Valid formats are: %s';
+			$msg = sprintf( $msgFormat, $format, '"' . implode( '", "', array_keys( self::$formats ) ) . '"' );
+			throw new NotFoundHttpException( $msg );
```
dayllanmaza, line 51: Why did you decide to use `NotFoundHttpException` and not something like InvalidArgumentException? Shouldn't an error caused by the client result in a 400?
In any case, if `WSExportInvalidArgumentException` is not in use anymore it should probably be removed

src/Controller/ExportController.php
```diff
@@ -136,6 +135,17 @@ private function getLang( Request $request ): string {
 		return strtolower( $lang );
 	}
 
+	/**
+	 * Get the format from the request.
+	 */
+	private function getFormat( Request $request ): string {
+		$format = $request->get( 'format' );
+		if ( !$format ) {
+			$format = 'epub-3';
```
dayllanmaza, line 144: I want to mention that `epub-3` is a different default than before. Making sure this is on purpose. 
Nit: Maybe update the method documentation to mention the default if `format` is not specified?

dayllanmaza (changes requested):

src/GeneratorSelector.php
samwilson: Yeah, I wondered about that, and really this change is not the most beautiful. I used a NotFoundHttpException because it's caught by the error-display system that we're already using for "revision not found" etc. Really, it'd probably be better to not have any HTTP stuff inside these classes, and keep all that in the controller; does that sound better?

src/Controller/ExportController.php
samwilson: It's on purpose; they're synonymous. Although that's not very obvious I guess!

Docblock updated.

src/GeneratorSelector.php
dayllanmaza: I would say so, yes. Also, `WSExportInvalidArgumentException` should be removed if moving away from it; I could not find any other uses besides this class

samwilson: Done. See what you think. It's slightly odd now, perhaps, in that it's duplicating the message — but I figure that that'll be sorted out when we add i18n to it all.

dayllanmaza (approved):

PR #288, opened 2020-12-10
Normalize element IDs to avoid duplicates or invalid IDs

samwilson:
Bug: T244694

### DIFF
--- a/src/PageParser.php
+++ b/src/PageParser.php
@@ -7,7 +7,12 @@ use DOMElement;
 use DOMXPath;
 
 class PageParser {
+
+	/** @var DOMXPath */
 	protected $xPath;
+
+	/** @var string[] Element ID values, used to fix duplicates. */
+	private static $ids = [];
 
 	/**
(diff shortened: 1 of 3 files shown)


src/PageParser.php
```diff
@@ -7,8 +7,13 @@
 use DOMXPath;
 
 class PageParser {
+
+	/** @var DOMXPath */
 	protected $xPath;
 
+	/** @var string[] Element ID values, used to fix duplicates. */
+	private static $ids = [];
```
cimurah, line 15: Looks good to me! Is there a reason why this array is static rather than dynamic?

samwilson: It's because we run PageParser on each separate chapter, but want a complete list of all IDs from all chapters, shared between all PageParser objects.

cimurah: I knew there was a strong reason for this, but couldn't figure out what it was. Thank you.

cimurah:
I've tested different books and the duplicate id warning is gone.

PR #294, opened 2020-12-18
Upgrade to Symfony 5.2

samwilson:


samwilson:
We can revisit this later if we want; no need for now.

PR #296, opened 2020-12-18
fix: support font embedding in Apple iBooks

JLHwung:
Apple Books requires this file for embedded fonts.

See https://help.apple.com/itc/booksassetguide/en.lproj/static.html#itc2cf4d26eb

also https://www.mobileread.com/forums/showthread.php?t=192389

Works on my local macOS and iOS:

image

Use case: I uses embed font to support characters in the Unicode SIP Plane.

### DIFF
--- a/src/Generator/EpubGenerator.php
+++ b/src/Generator/EpubGenerator.php
@@ -69,6 +69,7 @@ class EpubGenerator implements FormatGenerator {
 		$fileName = Util::buildTemporaryFileName( $book->title, 'epub' );
 		$zip = $this->createZipFile( $fileName );
 		$zip->addFromString( 'META-INF/container.xml', $this->getXmlContainer() );
+		$zip->addFromString( 'META-INF/com.apple.ibooks.display-options.xml', $this->getAppleIBooksDisplayOptionsXml() );
 		$zip->addFromString( 'OPS/content.opf', $this->getOpfContent( $book, $wsUrl ) );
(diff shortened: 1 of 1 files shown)


Tpt:
If I read https://help.apple.com/itc/booksassetguide/en.lproj/static.html#itc2cf4d26eb correctly, in ePub 3 the option should be set in the OPF file `<meta property="ibooks:specified-fonts">true</meta>` and there is no need for the new file. Wouldn't it be simpler?

JLHwung:
@Tpt I did try the ePub 3 properties mentioned in Apple docs. However it does not work for either macOS 11 or iOS 14. The ePub 2 trick in this PR works.

Tpt:
@JLHwung Thank you for checking it! Does ePub file with this change still work if they do not embed fonts?

JLHwung:
@Tpt That is a good question. It fallbacks to system font when the embed font is not found.

image

The "original" becomes an no-op but the ePub file is still working, at least in iOS/macOS. 😄 

Btw I realized that `FreeSerif` is always embedded when `lang` is `zh`.

https://github.com/wsexport/tool/blob/e55188beeada72519948a95fc059dc32517d3e98/src/Controller/ExportController.php#L122-L129

Not sure if the defaults here is required for `zh` as FreeSerif does not support Unified Ideographs. Guess we can just remove it and leave users configuring default fonts?

Tpt (approved):

Tpt:
Great! Indeed, it makes sense to stop FreeSerif for `zh` if it does not help. Feel free to do a PR for it.

PR #297, opened 2020-12-21
Enable heuristic processing for ebook-convert

dayllanmaza:
Some books take a very long time to convert to PDF. --enable-heuristics
speeds this process significantly

See: https://manual.calibre-ebook.com/conversion.html#heuristic-processing


samwilson (approved):
Works in my testing.

PR #298, opened 2020-12-22
Record book generation duration

samwilson:
Add a new 'duration' column to the DB table, to record the number
of seconds each book generation took.

Bug: T267079

### DIFF
--- a/src/Entity/GeneratedBook.php
+++ b/src/Entity/GeneratedBook.php
@@ -8,6 +8,7 @@ use DateTimeZone;
 // mediawiki-codesniffer apparently can't search annotations for unused classes.
 // phpcs:ignore MediaWiki.Classes.UnusedUseStatement.UnusedUse
 use Doctrine\ORM\Mapping as ORM;
+use Symfony\Component\Stopwatch\StopwatchEvent;
 
 /**
  * The GeneratedBook is the log of exports.
@@ -32,7 +33,7 @@ class GeneratedBook {
 
 	/**
 	 * @ORM\Column(type="datetime")
-	 * @var DateTime Time of the export.
+	 * @var DateTime Date and time of the export.
 	 */
 	private $time;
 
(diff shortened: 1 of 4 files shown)


src/Entity/GeneratedBook.php
```diff
@@ -128,4 +139,18 @@ public function getFormat(): string {
...
+
+	/**
+	 * Get the generation duration in seconds.
+	 * @return int|null
+	 */
+	public function getDuration(): ?int {
+		return $this->duration;
+	}
+
+	public function setDuration( ?int $duration ): self {
```
MusikAnimal, line 151: I'm guessing you used the Symfony CLI to create these methods when you added the column. Weird how it had it return `self` when it didn't for me! Not a big deal, but I would just make it `void` to be consistent with the other methods.

Also missing the doc block.

src/Entity/GeneratedBook.php
```diff
@@ -54,16 +55,26 @@ class GeneratedBook {
 	 */
 	private $format;
 
+	/**
+	 * @ORM\Column(type="integer", nullable=true)
+	 */
```
MusikAnimal, line 60: Suggest adding `@var StopwatchEvent|null $duration`

MusikAnimal (commented):
Looks good! Just a few very nitpicky comments.

MusikAnimal:
Oh, sorry didn't notice this was still WIP :)

src/Entity/GeneratedBook.php
samwilson: Done.

src/Entity/GeneratedBook.php
samwilson: Oh yeah, that is strange. Fixed both things.

samwilson (commented):
No worries, thanks for the speedy review!

PR #302, opened 2021-01-07
Stop using Guzzle's 'sink' option

samwilson:
There's a bug in the Guzzle cache middleware that means we have
to switch to saving the file responses (for images) in the response
handler rather than leaving it to Guzzle.

This is a workaround until the bug is fixed.

Bug: T265660

### DIFF
--- a/src/BookProvider.php
+++ b/src/BookProvider.php
@@ -213,15 +213,21 @@ class BookProvider {
 		$requests = function () use ( $client, $pictures, $cache ) {
 			foreach ( $pictures as $picture ) {
 				$url = $picture->url;
-				$tempFile = $cache->getDirectory() . '/' . uniqid( 'pic-' );
-				$picture->tempFile = $tempFile;
-				yield function () use ( $client, $url, $tempFile ) {
-					return $client->getAsync( $url, [ 'sink' => $tempFile ] );
+				yield function () use ( $client, $url ) {
+					// We could use the 'sink' option here, but for https://github.com/Kevinrob/guzzle-cache-middleware/issues/82
+					return $client->getAsync( $url );
 				};
 			}
 		};
 		$pool = new Pool( $client, $requests(), [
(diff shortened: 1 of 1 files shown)


src/BookProvider.php
```diff
@@ -213,23 +213,28 @@ protected function getPicturesData( array $pictures ) {
 		$requests = function () use ( $client, $pictures, $cache ) {
 			foreach ( $pictures as $picture ) {
 				$url = $picture->url;
-				$tempFile = $cache->getDirectory() . '/' . uniqid( 'pic-' );
-				$picture->tempFile = $tempFile;
-				yield function () use ( $client, $url, $tempFile ) {
-					return $client->getAsync( $url, [ 'sink' => $tempFile ] );
+				yield function () use ( $client, $url ) {
+					// We could use the 'sink' option here, but for https://github.com/Kevinrob/guzzle-cache-middleware/issues/82
```
cimurah, line 217: Looks good and works as expected :) 
I looked at the issue posted on [github](https://github.com/Kevinrob/guzzle-cache-middleware/issues/82) and it was posted on 2017. It made me wonder if this hasn't been fixed since then. I came across https://stackoverflow.com/questions/57578626/guzzle-6-download-file/57606637#57606637. I'm wondering if the `sink `option is coming back empty because we need to pass a resource rather than the file path as stated in the [docs](https://docs.guzzlephp.org/en/latest/request-options.html#sink). I was going to test out this theory but you might have tried it already. I think it is ready to merge otherwise.

samwilson: Good find! I hadn't checked that. But no, it seems it's the same issue: the cached response produces a zero-sized temp `pic-` file. I think reproducing the cached file on each run is the only workaround (which isn't as inefficient as it sounds, because surely that's exactly what we'd want the cache middleware to do for us anyway?).

cimurah: Makes sense. Thank you!

PR #307, opened 2021-01-11
README: instruct to run all migrations, not just creating the db

MusikAnimal:
Bug: T266190

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,17 +39,23 @@ Installation
 3. Create a mysql database and database user
    and add these details to `.env.local`.
 
-4. Run `./bin/console doctrine:database:create` to initialize the database.
+4. Create the database with `./bin/console doctrine:database:create`
+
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -92,6 +92,7 @@
 		],
 		"test": [
 			"composer validate",
+			"./bin/console doctrine:migrations:migrate --env=test --no-interaction",
```
samwilson, line 95: It doesn't really matter because this runs pretty quickly, but this feels like it should be paired with a Composer script that runs phpunit (or that should be added here). As it is, this command isn't required to run the rest of `test`, but we'd be assuming people would run `composer test` before `./bin/phpunit` (which they probably would, so this doesn't really matter...).

MusikAnimal: Yeah I added it here thinking "why not". It indeed doesn't really make sense given this script does not run the tests that depend on the migrations, as you say :)

I'll just remove it. The README changes are enough for now, I think.

PR #308, opened 2021-01-12
Add feature flag for caching

samwilson:
Add APP_ENABLE_CACHE environment variable, defaulting to false,
to enable caching of API requests.

Bug: T271771


samwilson:
This doesn't implement this 100% as we should do (e.g. the command still lists `nocache` as an option even when caching is turned off), because I think we'll want to just remove it in a month or so. I think it fulfills what we need it for.

PR #309, opened 2021-01-13
Cache 'about' page with app cache

samwilson:
Switch from manually storing the about pages in files, to using
Symfony cache with a TTL of 1 month.

Bug: T271876


MusikAnimal (approved):
Works in my testing

PR #310, opened 2021-01-13
Exclude interwiki links from chapter lists

samwilson:
Bug: T270395


MusikAnimal (approved):
This fixes the bug for the examples given at https://phabricator.wikimedia.org/T270395 and code looks good to me.

PR #316, opened 2021-01-19
Transliterate element IDs and fix replacement process

samwilson:
The existing system was still creating duplicates where there were
no ascii characters in the original ID (they were all replaced by
underscores, so only had to match on the length).

Also switched from appending a count of each duplicate ID to just
using a count of all IDs in order to create a unique ID, because
array_count_values() isn't very fast and we don't actually care
about this number.

Bug: T272119

### DIFF
--- a/src/PageParser.php
+++ b/src/PageParser.php
@@ -295,23 +295,30 @@ class PageParser {
 			$id = $oldId;
 			// Check for duplicates and fix them by appending a count.
 			if ( array_search( $id, self::$ids ) !== false ) {
-				$id .= '-n' . ( array_count_values( self::$ids )[ $id ] + 1 );
+				$id .= '-n' . ( count( self::$ids ) + 1 );
 			}
+			// Transliterate ID and ensure it contains only certain characters.
+			$idTrans = transliterator_transliterate( 'Any-Latin; Latin-ASCII; ', $id );
+			$idTransConv = iconv( 'UTF-8', 'ASCII//TRANSLIT//IGNORE', $idTrans );
+			$id = preg_replace( '/[^a-zA-Z0-9\-:.]/', '_', $idTransConv );
 			// Must start with a letter.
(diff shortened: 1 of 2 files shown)


src/PageParser.php
```diff
@@ -294,21 +294,28 @@ protected function cleanIds(): void {
 			if ( array_search( $id, self::$ids ) !== false ) {
 				$id .= '-n' . ( array_count_values( self::$ids )[ $id ] + 1 );
 			}
+			// Transliterate ID and ensure it contains only certain characters.
+			$idTrans = transliterator_transliterate( 'Any-Latin; Latin-ASCII; ', $id );
+			$idTransConv = iconv( 'UTF-8', 'ASCII//TRANSLIT//IGNORE', $idTrans );
+			$id = preg_replace( '/[^a-zA-Z0-9\-_:.]/', '_', str_replace( ' ', '_', $idTransConv ) );
```
Tpt, line 300: I don't see why the `str_replace` is useful.

Tpt (approved):
Looks good to me after a quick read (I have not tested anything).

src/PageParser.php
samwilson: Good point. Neither do I! :-) Removed.

cimurah:
Looks good! I've tested this and I don't get the `Error: Maximum execution time of 30 seconds exceeded` error anymore. I tried the book `format=epub&lang=it&page=Divina_Commedia` and still getting duplicates id messages when I run epub check in my computer. It looks like there is a bunch of `item` tags that have an `id` but are not being picked up by the `cleanIds()` function.

samwilson:
It looks like that's a separate issue: I think what's happening is that the subpages are being listed on the main top-level ToC as well as on the subpage ToCs (the three main parts), and so they're being duplicated in the book. Should be fixed of course, but I think in a separate patch.

PR #317, opened 2021-01-19
Add app:check command for easier running of epubcheck

samwilson:


MusikAnimal (approved):
This seems like a very powerful utility for debugging! Amazing stuff! I didn't test every option but I'm sure they all work. I take it if you don't supply any arguments, it runs on a bunch of test books? Which is really neat :) Maybe the behaviour could be explained in the --help output?

That's my only comment, otherwise looks +1 to me :)

samwilson:
> This seems like a very powerful utility for debugging! Amazing stuff! I didn't test every option but I'm sure they all work. I take it if you don't supply any arguments, it runs on a bunch of test books? Which is really neat :) Maybe the behaviour could be explained in the --help output?
> 
> That's my only comment, otherwise looks +1 to me :)

Thanks! And yeah, that's a good idea. I've updated the help text a bit, see what you think.

PR #318, opened 2021-01-21
Fix epubcheck tests to only report failures

samwilson:
* Move the epubcheck classes from tests to src in preparation
  for using them in other places.
* Switch to test books (on en and fr) that don't contain errors.
* Fix CSS test to use valid CSS.
* Include once-failing tests in CI.

Bug: T268000


cimurah:
If I understand correctly, are supposed to see some failing errors? 

I get 4 failures in this patch when I run `./bin/phpunit`.

samwilson:
What version of epubcheck do you have? The latest is 4.2.4, and things change quite a lot between versions as far as support for newer HTML goes.

I just realised that some of the failing EPUB 2 tests that were failing under epubcheck 4.2.2 are actually fine in the later version (which is good, because Calibre *should* be producing valid EPUB 2). So I've removed the `exclude-from-ci` phpunit group all together and things seem fine.

samwilson:
No, scratch that, I was testing with a bad cache. The EPUB 2 tests are still failing.

samwilson:
@dayllanmaza the version of epubcheck in Docker is out of date. Should we do something like this patch is doing for CI? (i.e. install the latest version manually) I'm not sure of the proper way to do that in Docker.

samwilson:
The failures here are being caused by this duplicate IDs, fixed in #325.

samwilson:
Some of the other failures here (also about duplicate IDs) will be fixed by #322.

samwilson:
I've temporarily added what I think are the required fixes from #322 to this branch, and everything is now passing.

samwilson:
This is ready for review now. (Removed the temp changes.)

PR #321, opened 2021-01-27
Don't remove child nodes of figures and images.

samwilson:
When processing img elements, this was replacing figure and
figure-inline with an img element. This was discarding any
figcaption or other contents of the node.

It doesn't look like there's any reason to do this replacement,
as we're already modifying the img element to our requirements.

Also switch to ctype_xdigit to check for the thumb URL segments
being valid hex digits, because using is_numeric to do this was
deprecated as of PHP 7.0.

Also check for any contained img elements in figure etc. and break
if none found.

Bug: https://phabricator.wikimedia.org/T272307
Bug: https://phabricator.wikimedia.org/T272180

### DIFF
--- a/src/PageParser.php
+++ b/src/PageParser.php
@@ -144,14 +144,17 @@ class PageParser {
 	}
 
 	/**
-	 * return the pictures of the file, for all handled <img, the alt
-	 * attribute is set to the title of the image so the backend can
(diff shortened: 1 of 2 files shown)


samwilson:
Rebased.

MusikAnimal (approved):
Works in my testing

PR #322, opened 2021-02-01
Fix references sometimes opening as external links

dayllanmaza:
I'm unsure about the repercussions of changing how we load the html
content from XML to HTML but the later seems to fix this issue. It
is also how we load pages' content from parsoid.

This issue was happening becuase `$this->xPath->query( '//html:a[contains(@style, "mw-Ref")]' )`
was not working when the DOMDocument content was loaded via loadHTML.
Removing the namespace from the above query fixes the book references but breaks the About
page unless it is loaded as HTML.

Bug: T270372

### DIFF
--- a/src/BookProvider.php
+++ b/src/BookProvider.php
@@ -3,6 +3,7 @@
 namespace App;
 
 use App\Util\Api;
+use App\Util\Util;
 use DOMDocument;
 use GuzzleHttp\Pool;
 use GuzzleHttp\Promise\PromiseInterface;
@@ -182,12 +183,7 @@ class BookProvider {
 	}
 
 	protected function domDocumentFromHtml( $html ) {
-		$document = new DOMDocument( '1.0', 'UTF-8' );
-		libxml_use_internal_errors( true );
-		$document->loadHTML( mb_convert_encoding( str_replace( '<?xml version="1.0" encoding="UTF-8" ?>', '', $html ), 'HTML-ENTITIES', 'UTF-8' ) );
-		libxml_clear_errors();
-		$document->encoding = 'UTF-8';
-		return $document;
+		return Util::buildDOMDocumentFromHtml( $html );
 	}
 
 	/**
--- a/src/PageParser.php
+++ b/src/PageParser.php
@@ -418,15 +418,9 @@ class PageParser {
 	}
 
 	private function cleanReferenceLinks() {
-		// Get all links that contain the "style" value "mw-Ref"
-		$links = $this->xPath->query( '//html:a[contains(@style, "mw-Ref")]' );
-		foreach ( $links as $link ) {
-			$href = $link->getAttribute( 'href' );
-			$pos = strpos( $href, '#' );
-			$link->setAttribute( 'href', substr( $href, $pos ) );
-		}
-		// Get all links that have the "rel" attribute equals to "mw-Ref"
-		$links = $this->xPath->query( '//html:a[@rel="mw:referencedBy"]' );
+		$links = $this->xPath->query(
+			'//*[@typeof="mw:Extension/ref"]/a | //a[@rel="mw:referencedBy"]'
+		);
 		foreach ( $links as $link ) {
 			$href = $link->getAttribute( 'href' );
 			$pos = strpos( $href, '#' );
--- a/src/Util/Api.php
+++ b/src/Util/Api.php
@@ -4,7 +4,6 @@ namespace App\Util;
 
 use App\PageParser;
 use DateInterval;
-use DOMDocument;
 use Exception;
 use GuzzleHttp\Client;
 use GuzzleHttp\ClientInterface;
@@ -145,9 +144,7 @@ class Api {
 				return $oldWikisourceApi->getPageAsync( 'MediaWiki:Wsexport_about' )->wait();
 			}
 		} );
-		// Rewrite some parts of the returned HTML.
-		$document = new DOMDocument( '1.0', 'UTF-8' );
-		$document->loadXML( $content );
+		$document = Util::buildDOMDocumentFromHtml( $content );
 		$parser = new PageParser( $document );
 		$document = $parser->getContent( true );
 		// Add https to protocol-relative links.
--- a/src/Util/Util.php
+++ b/src/Util/Util.php
@@ -4,6 +4,7 @@ namespace App\Util;
 
 use App\FileCache;
 use App\Refresh;
+use DOMDocument;
 use DOMElement;
 use DOMXPath;
 use Exception;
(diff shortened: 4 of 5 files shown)


dayllanmaza:
I'm not very familiar with any of this so I need some help here making sure this patch does not break something else. In my tests so far it seems to work

dayllanmaza:
Actually it does not fix this for some of the bugs in the task

samwilson:
Switching to loadHTML() sounds like it should be fine. I guess there's other ways we're changing links though, that aren't all being caught (or too many of them are).

src/PageParser.php
```diff
@@ -418,15 +418,9 @@ private function cleanRedLinks() {
...
-		$links = $this->xPath->query( '//html:a[contains(@style, "mw-Ref")]' );
-		foreach ( $links as $link ) {
-			$href = $link->getAttribute( 'href' );
-			$pos = strpos( $href, '#' );
-			$link->setAttribute( 'href', substr( $href, $pos ) );
-		}
-		// Get all links that have the "rel" attribute equals to "mw-Ref"
-		$links = $this->xPath->query( '//html:a[@rel="mw:referencedBy"]' );
+		$links = $this->xPath->query(
+			'//a[contains(@style, "mw-Ref") or @rel="mw:referencedBy" or @rel="mw:WikiLink"]'
```
dayllanmaza, line 422: As it turns out, links for some of the examples in the task did not match `a[contains(@style, "mw-Ref") or @rel="mw:referencedBy"` and instead they had  `@rel="mw:WikiLink"`

Does anyone knows why they are different depending on the source? I'm afraid we could be missing other variations

samwilson: I bet there are things we're missing. :-) I wouldn't worry about catching absolutely everything though; we can follow up with more.

Are they actual reference links? Or just normal links? Or normal links with a reference? (Which page are these on?)

src/Util/Api.php
```diff
@@ -147,7 +147,7 @@ public function getAboutPage(): string {
 		} );
 		// Rewrite some parts of the returned HTML.
 		$document = new DOMDocument( '1.0', 'UTF-8' );
-		$document->loadXML( $content );
+		$document->loadHTML( $content );
```
samwilson, line 150: I'm getting this epubcheck error with this change:

> ERROR(RSC-005): book.epub/OPS/about.xhtml(-1,-1): Error while parsing file: The processing instruction target matching "[xX][mM][lL]" is not allowed.

I think the XML prolog has to be stripped first.

dayllanmaza: Thanks for finding this. I'll take a look

src/PageParser.php
dayllanmaza: The ones in here https://wsexport.wmflabs.org/?lang=en&page=The_Variation_of_Animals_and_Plants_under_Domestication%2FXIV&format=epub-3&fonts=

I'm not sure I know the difference between a `reference link` and `normal links with a reference`. I guess it is either of those?

DenisGagne52: Reference links are those built in Ws with ref tags to create footnotes. I believe they were catched with the cleanReferenceLinks() function.

I think that normal links are catched with the setlinks() function but it doesn't work anymore when the full path is not provided -> in this situation :  if ( substr( $href, 0, 1 ) === '#' )  ; but it does work for all the  elseif situations.

Some references are hand-built and may appear as normal links.

src/Util/Api.php
samwilson: The PageParser changes work fine for me without changing this line.

src/PageParser.php
samwilson: I think the xpath to select all reference links and backlinks should be something like:

    //*[@typeof="mw:Extension/ref"]/a | //a[@rel="mw:referencedBy"]

dayllanmaza: Thank you @DenisGagne52.

@samwilson that query works in most cases that I've tested but this one sleeps thru 
`<sup id="ref_136" class="plainlinks" about="#mwt7" typeof="mw:Transclusion"><a rel="mw:WikiLink" href="#endnote_136">[136]</a></sup>` taken from https://wsexport.wmflabs.org/?lang=en&page=The_Variation_of_Animals_and_Plants_under_Domestication%2FXIV&format=epub-3&fonts=

I've updated the patch accordingly

src/Util/Api.php
dayllanmaza: > The PageParser changes work fine for me without changing this line.

I'm not sure what you mean.

src/PageParser.php
samwilson: That page is constructing 'manual' footnote links, with e.g. `[[#endnote_123|[123]]]` (and matching backlinks from the footnotes, constructed in a similar manner). This means that there's nothing particular about the links that we can match on — the `<sup typeof="mw:Transclusion">` is just a coincidence because it's in a template.

I think what we need to do is strip the page title, which is not currently available in the PageParser class. We'll have to pass it in everywhere.

A link such as `[[#Foo]]` on page `Lorem` (it doesn't matter if it has a leading pagename or alternate display text) is rendered as `<a href="./Lorem#Foo">#Foo</a>`, so we need to look for `./Lorem#` and remove everything before `#` (but not if it's e.g. `[[Lorem/Ipsum#Foo]]` because that's a different page).

I set up some tests at https://en.wikisource.beta.wmflabs.org/wiki/Links

src/BookProvider.php
```diff
@@ -182,12 +183,7 @@ protected function getDocument( $title ) {
 	}
 
 	protected function domDocumentFromHtml( $html ) {
```
samwilson, line 185: Probably don't really need this method any more.

src/PageParser.php
```diff
@@ -418,15 +418,9 @@ private function cleanRedLinks() {
...
-		$links = $this->xPath->query( '//html:a[contains(@style, "mw-Ref")]' );
-		foreach ( $links as $link ) {
-			$href = $link->getAttribute( 'href' );
-			$pos = strpos( $href, '#' );
-			$link->setAttribute( 'href', substr( $href, $pos ) );
-		}
-		// Get all links that have the "rel" attribute equals to "mw-Ref"
-		$links = $this->xPath->query( '//html:a[@rel="mw:referencedBy"]' );
+		$links = $this->xPath->query(
+			'//*[@typeof="mw:Extension/ref"]/a | //a[@rel="mw:referencedBy"]'
```
samwilson, line 422: The `*` can probably safely be `sup`, although it probably isn't massively different performance-wise.

samwilson (commented):
Looks good, and works in my testing. I do wonder if we want to add tests for the references HTML (which we can later extend when we deal with the 'manual references' stuff if Parsoid doesn't sort it out for us before then).

PR #323, opened 2021-02-01
Override for 'nan' language code → 'zh-min-nan'

samwilson:
Add special handling for the one Wikisource whose language code
doesn't match its subdomain.

Bug: https://phabricator.wikimedia.org/T272043


MusikAnimal (approved):
All things considered this seems like a suitable solution. We can revisit later if there's any pushback about the `xml:lang="zh-min-nan"` bit.

PR #324, opened 2021-02-02
Upgrade PHP dependencies, including the ToolforgeBundle

samwilson:


MusikAnimal:
This is :+1: but maybe let's get https://github.com/wikimedia/ToolforgeBundle/pull/42 merged first, push out another release then change this to use that version. Otherwise Docker peeps can't make use of the ReplicasClient and SshCommand

samwilson:
> This is +1 but maybe let's get [wikimedia/ToolforgeBundle#42](https://github.com/wikimedia/ToolforgeBundle/pull/42) merged first, push out another release then change this to use that version. Otherwise Docker peeps can't make use of the ReplicasClient and SshCommand

Good idea.

PR #331, opened 2021-02-15
Add on-wiki config to define default fonts

samwilson:
Set up an on-wiki config page MediaWiki:WS_Export.json, which
can be used to store any per-Wikisource configuration. So far
this is just `defaultFont` but it can be extended as needed.

Defaults for these configs live in the tool's code, not in
Multilingual Wikisource (like the About page) because that wiki
will need its own config.

The existing default of FreeSerif will need to be migrated to
the on-wiki config once this change has been merged.

Bug: https://phabricator.wikimedia.org/T274561

### DIFF
--- a/tests/Http/BookTest.php
+++ b/tests/Http/BookTest.php
@@ -100,4 +100,20 @@ class BookTest extends WebTestCase {
 		$this->assertStringContainsString( '<input name="page" id="page" type="text" size="30" required="required" class="form-control"
 			value="A &quot;title&quot;" />', $client->getResponse()->getContent() );
 	}
+
+	public function testFonts() {
+		$client = static::createClient();
+		// No font.
+		$client->request( 'GET', '/', [ 'fonts' => 'Not a font name' ] );
+		$this->assertStringContainsString( '<option value="" selected="selected">None (use device default)</option>', $client->getResponse()->getContent() );
+		// Default when there's no on-wiki config.
+		$client->request( 'GET', '/', [ 'lang' => 'en' ] );
+		$this->assertStringContainsString( '<option value="" selected="selected">None (use device default)</option>', $client->getResponse()->getContent() );
+		// Default font from on-wiki config.
+		$client->request( 'GET', '/', [ 'lang' => 'beta' ] );
+		$this->assertStringContainsString( '<option value="DejaVu&#x20;Sans&#x20;Mono" selected="selected">', $client->getResponse()->getContent() );
+		// Default is overridden in request.
+		$client->request( 'GET', '/', [ 'lang' => 'beta', 'fonts' => 'DejaVu Serif' ] );
+		$this->assertStringContainsString( '<option value="DejaVu&#x20;Serif" selected="selected">', $client->getResponse()->getContent() );
+	}
 }
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -61,5 +61,6 @@
 	"error-page-header": "An error occurred",
 	"error-page-details": "Error details: $1",
 	"error-page-issue": "If this error persists, please $1.",
-	"error-page-issue-link": "report an issue on Phabricator"
+	"error-page-issue-link": "report an issue on Phabricator",
+	"onwikiconfig-failure": "Unable to retrieve the on-wiki configuration from: $1"
 }
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -63,5 +63,6 @@
 	"error-page-header": "Header text for the generic error page.",
 	"error-page-details": "Introduction text to the generic error page.\n\n$1 - Error description.",
 	"error-page-issue": "Call to action text.\n\n$1 - HTML link, with text from error-page-issue-link below.",
-	"error-page-issue-link": "Link text to use in the link in error-page-issue above."
+	"error-page-issue-link": "Link text to use in the link in error-page-issue above.",
+	"onwikiconfig-failure": "Error message displayed if the on-wiki configuration fetching failed.\n\n* $1 - the URL of the config JSON page on Wikisource."
 }
--- a/src/Controller/ExportController.php
+++ b/src/Controller/ExportController.php
@@ -106,7 +106,7 @@ class ExportController extends AbstractController {
 			}
 		}
 
-		$font = $this->getFont( $request, $api->getLang(), $fontProvider );
+		$font = $this->getFont( $request, $fontProvider );
 		$credits = (bool)$request->get( 'credits', true );
 		$images = (bool)$request->get( 'images', true );
 		return $this->render( 'export.html.twig', [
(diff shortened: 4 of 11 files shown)


src/Util/Api.php
```diff
@@ -95,6 +98,30 @@ public function getLang(): string {
...
+		if ( isset( $this->wikiConfig[ $this->getLang() ] ) ) {
+			return $this->wikiConfig[ $this->getLang() ];
+		}
+		$this->wikiConfig[ $this->getLang() ] = $this->cache->get( 'wikiConfig_' . $this->getLang(), function ( CacheItemInterface $cacheItem ) {
+			$cacheItem->expiresAfter( new DateInterval( 'P1M' ) );
+			$configPageName = 'MediaWiki:WS_Export.json';
+			$json = $this->get( 'https://' . $this->getDomainName() . '/w/index.php?title=' . $configPageName . '&action=raw&ctype=application/json' );
+			$data = json_decode( $json, true );
+			if ( $data === null ) {
+				return [];
```
dayllanmaza, line 115: Are we OK breaking backward compatibility by removing the defaults for non-latin languages? Should we provide some defaults?

src/Util/Api.php
```diff
@@ -95,6 +98,30 @@ public function getLang(): string {
 		return $this->lang;
 	}
 
+	/**
+	 * Get the on-wiki configuration from the `MediaWiki:WS_Export.json` page. Cached for 1 month.
+	 * @return mixed[] With keys: defaultFont
+	 */
+	public function getWikiConfig(): array {
```
dayllanmaza, line 105: I feel like this methods belongs in its own service. Something like `WikisourceConfig` or `RemoteConfig` perhaps? What do you think?

Also, should we protect against `getDomainName()` or `getLang()` being empty?

src/Util/Api.php
samwilson: I mentioned it in the commit message: we'll migrate these defaults to the on-wiki configs on the respective wikis. There might be a short time when things don't work correctly, but not long.

src/Util/Api.php
samwilson: Good idea, then that service can take care of the structure of the page too, instead of having it spread out wherever it's used. I'll move it.

tests/Http/BookTest.php
```diff
@@ -94,4 +94,20 @@ public function testTitlePrefill() {
...
 					value="A &quot;title&quot;" />', $client->getResponse()->getContent() );
 	}
+
+	public function testFonts() {
+		$client = static::createClient();
+		// No font.
+		$client->request( 'GET', '/', [ 'fonts' => 'Not a font name' ] );
+		$this->assertStringContainsString( '<option value="" selected="selected">None (use device default)</option>', $client->getResponse()->getContent() );
+		// Default when there's no on-wiki config.
+		$client->request( 'GET', '/', [ 'lang' => 'en' ] );
```
dayllanmaza, line 104: Isn't this gonna fail when the default font is set on enwikisource?

dayllanmaza (changes requested):
Should we add tests for OnWikiConfig? Other than the comments below, it looks good to me.

tests/Http/BookTest.php
samwilson: No, the idea is that if you give it an invalid font name it *doesn't* fall back to the default, it just doesn't embed one. Should it use the default in that situation? Also, I think it should be possible to choose *not* to embed a font (which hasn't been the case for non-latin languages); that's a separate issue I think.

src/Util/Api.php
```diff
@@ -49,6 +49,9 @@ class Api {
 	/** @var string[][] */
 	private $namespaces = [];
 
+	/** @var mixed[] */
+	private $wikiConfig;
```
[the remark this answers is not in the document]
samwilson: Fixed.

Should we add Phan? It'd catch that I think.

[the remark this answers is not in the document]
samwilson: I created https://phabricator.wikimedia.org/T276519 for Phan.

tests/Http/BookTest.php
dayllanmaza: What I mean is that on line 103 it will go and fetch the defaultFont on en.wikisource.org. If I create that json file right now, the assert on 104 will fail because it will actually find a default font. Does that makes sense? Just like with beta, if I change the defaultFont there line 106 will start failing. Am I missing something?

Everything else looks great btw

samwilson: That's a good point.

I sort of think this isn't too much of an issue because a) it's unlikely that English Wikisource will set a default font; and b) this whole test class is based on on-wiki data that might change.

The trouble here is that to mock this isn't too straight forward, because this is meant to be the highest level of integration tests (that test the whole stack). I've tried mocking and changing the test client's container, but it doesn't work:

```php
		$client = static::createClient();
		$onWikiConfig = $this->createMock( OnWikiConfig::class );
		$onWikiConfig->expects( $this->once() )
			->method( 'getDefaultFont' )
			->willReturn( 'foo' );
		$client->getContainer()->set( OnWikiConfig::class, $onWikiConfig );
```

From what I read, this is generally considered a bad idea.

Any ideas?

dayllanmaza: I just tried it locally and I was able to run the test with some minor changes. [Here is what I tried](https://github.com/wikimedia/ws-export/files/6172784/patch.txt)

Why is it considered a bad idea? Typically, you wouldn't mock services when running integration tests but external resources that we can't control are an exception IMO unless you can guarantee that the external config is not gonna change. @cimurah, @MusikAnimal, any thoughts?

For an even more "stricter" test you could mock API response instead of OnWikiConfig

In any case, the rest of the patch is good to go and I'm happy to merge it if you want to keep it as it is.

cimurah: I really don't have much to say since tests can still be confusing for me.  I need more experience around mocking.

> Why is it considered a bad idea? Typically, you wouldn't mock services when running integration tests but external resources that we can't control are an exception IMO unless you can guarantee that the external config is not gonna change. @cimurah, @MusikAnimal, any thoughts?

I do want to add that reading this review alone helped a lot. I understand what @dayllanmaza meant with the example he provided, but not sure if it's necessary for this case.

samwilson: @dayllanmaza That's pretty much what I was trying too, but it doesn't work. Try changing your code to `->will( $this->onConsecutiveCalls( 'foo', 'bar', 'baz', 'bif' ) );` and you'll see the tests still pass — they're still using the actual OnWikiConfig service and making the call to the API.

As for why it's a bad idea, I'm sort of on the fence about it, but I think the idea is that functional tests should be thought of as high-level integration tests, and that it's more about making sure everything works together and treading the system itself as a black box. Anyway, I'd be happy to do it, but can't figure out how to get it working.

PR #332, opened 2021-02-17
Fix for RTL display

samwilson:
Add lang and dir params to `<html>`, and fix column display.

Bug: T274957

### DIFF
--- a/templates/base.html.twig
+++ b/templates/base.html.twig
@@ -1,9 +1,12 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="{{ lang() }}" dir="{% if is_rtl() %}rtl{% else %}ltr{% endif %}">
 <head>
 	<meta charset="UTF-8"/>
 	<title>{{ msg('app-title') }}</title>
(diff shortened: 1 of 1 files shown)


templates/base.html.twig
```diff
@@ -27,6 +27,13 @@
 			margin: -20px -20px 20px;
 		}
 
+		{% if is_rtl() %}
+		{# Hacky fix for RTL with Bootstrap 3. #}
+		.col-lg-1, .col-lg-2, .col-lg-3, .col-lg-4, .col-lg-5, .col-lg-6, .col-lg-7, .col-lg-8, .col-lg-9, .col-lg-10, .col-lg-11, .col-lg-12 {
+			float: right;
+		}
```
MusikAnimal, line 34: you could also do:
```
{% if is_rtl() %}
<link rel="stylesheet" type="text/css" href="https://tools-static.wmflabs.org/cdnjs/ajax/libs/bootstrap-rtl/3.4.0/css/bootstrap-flipped.min.css">
{% endif %}
```
which should cover all of Bootstrap. But with as simple as WS Export is I think this 'hack' is probably sufficient.

MusikAnimal (approved):
:+1:

templates/base.html.twig
samwilson: Thank you! I didn't know that existed.

PR #333, opened 2021-02-18
Improve ID parsing

samwilson:
* Remove Parsoid-style ID attributes.
* Move fragment substring checking into PHP instead of XPath, where it's quicker.

Bug: https://phabricator.wikimedia.org/T272119


MusikAnimal (approved):
I can confirm with this code exporting https://de.wikisource.org/wiki/Die_Gartenlaube does not time out anymore. One more Parsoid issue out of the way :+1:

PR #334, opened 2021-02-21
Add list of Wikisource languages to form

samwilson:
Add a dropdown to the main export form to display all available
Wikisource langauges, as well as whatever is specified by the
?lang= query string parameter.

Bug: https://phabricator.wikimedia.org/T269344


cimurah:
This looks good to me and works in my testing! I just have a very very tiny nitpick. For multilingual wikisource, we see `Multilingual Wikisource ()` as the option. We would need to create a new label when the language doesn't apply or we can leave it as it. What do you think?

samwilson:
> This looks good to me and works in my testing! I just have a very very tiny nitpick. For multilingual wikisource, we see `Multilingual Wikisource ()` as the option. We would need to create a new label when the language doesn't apply or we can leave it as it. What do you think?

Good catch! It should probably say `(mul)` because https://mul.wikisource.org is an alias for that Wikisource. I'll update the patch.

src/Wikidata.php
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	/**
+	 * Get the XML result of a Sparql query.
+	 * @param string $query The Sparql query to execute.
+	 * @return DOMDocument
+	 */
+	protected function getXml( $query ) {
+		$url = "https://query.wikidata.org/bigdata/namespace/wdq/sparql?query=" . urlencode( $query );
+		$client = new Client();
```
dayllanmaza, line 81: Sorry for the late review. I think this can be injected. I've submitted a follow up patch here https://github.com/wikimedia/ws-export/pull/339

Let me know what you think

PR #335, opened 2021-02-23
Remove phetools dependency

cimurah:
Remove credits phetools api dependency by quering the corresponding tables to
get the book credits directly in ws export.

Bug: T257886

### DIFF
--- a/.env
+++ b/.env
@@ -42,3 +42,24 @@ APP_LOG_RECIPIENT_1=admin1@example.org
 APP_LOG_RECIPIENT_2=admin2@example.org
 APP_LOG_SUBJECT="[Wikisource Export]"
 EPUBCHECK_JAR="/usr/bin/epubcheck"
+# For production, REPLICAS_HOST should be *.web.db.svc.eqiad.wmflabs (or analytics instead of web)
+# For local environments, use "127.0.0.1" and "host.docker.internal" for docker.
+REPLICAS_HOST_S1="127.0.0.1"
+REPLICAS_HOST_S2="127.0.0.1"
+REPLICAS_HOST_S3="127.0.0.1"
+REPLICAS_HOST_S4="127.0.0.1"
+REPLICAS_HOST_S5="127.0.0.1"
+REPLICAS_HOST_S6="127.0.0.1"
+REPLICAS_HOST_S7="127.0.0.1"
+REPLICAS_HOST_S8="127.0.0.1"
+#For local environments, use any safe range of ports (such as 4711 for s1, 4712 for s2, and so on)
+REPLICAS_PORT_S1=4711
+REPLICAS_PORT_S2=4712
+REPLICAS_PORT_S3=4713
+REPLICAS_PORT_S4=4714
+REPLICAS_PORT_S5=4715
+REPLICAS_PORT_S6=4716
+REPLICAS_PORT_S7=4717
+REPLICAS_PORT_S8=4718
+REPLICAS_USERNAME="username"
+REPLICAS_PASSWORD="password123"
--- a/README.md
+++ b/README.md
@@ -43,6 +43,18 @@ Installation
 
 5. Run the migrations with `./bin/console doctrine:migrations:migrate`
 
+6. This tool uses the [Toolforge Bundle](https://github.com/wikimedia/ToolforgeBundle), and it connects to [multiple databases](https://github.com/wikimedia/ToolforgeBundle#replicas-connection-manager).
+  * Set replicas credentials in the `.env.local` file.
+
+  * Establish an SSH tunnel to the replicas (only necessary on local environments)
+```bash
+$ ./php bin/console toolforge:ssh
+```
+  * Bind address for docker enviroments
+```bash
+$ php bin/console toolforge:ssh --bind-address=0.0.0.0
+```
+
 Tests
 =====
 
--- a/config/packages/doctrine.yaml
+++ b/config/packages/doctrine.yaml
@@ -5,6 +5,46 @@ doctrine:
                 url: '%env(resolve:DATABASE_URL)%'
                 driver: pdo_mysql
                 charset: utf8mb4
+            toolforge_s1:
+                host: '%env(REPLICAS_HOST_S1)%'
+                port: '%env(REPLICAS_PORT_S1)%'
+                user: '%env(REPLICAS_USERNAME)%'
+                password: '%env(REPLICAS_PASSWORD)%'
+            toolforge_s2:
+                host: '%env(REPLICAS_HOST_S2)%'
+                port: '%env(REPLICAS_PORT_S2)%'
+                user: '%env(REPLICAS_USERNAME)%'
+                password: '%env(REPLICAS_PASSWORD)%'
+            toolforge_s3:
+                host: '%env(REPLICAS_HOST_S3)%'
+                port: '%env(REPLICAS_PORT_S3)%'
+                user: '%env(REPLICAS_USERNAME)%'
+                password: '%env(REPLICAS_PASSWORD)%'
+            toolforge_s4:
+                host: '%env(REPLICAS_HOST_S4)%'
+                port: '%env(REPLICAS_PORT_S4)%'
+                user: '%env(REPLICAS_USERNAME)%'
+                password: '%env(REPLICAS_PASSWORD)%'
+            toolforge_s5:
+                host: '%env(REPLICAS_HOST_S5)%'
+                port: '%env(REPLICAS_PORT_S5)%'
+                user: '%env(REPLICAS_USERNAME)%'
+                password: '%env(REPLICAS_PASSWORD)%'
+            toolforge_s6:
+                host: '%env(REPLICAS_HOST_S6)%'
+                port: '%env(REPLICAS_PORT_S6)%'
+                user: '%env(REPLICAS_USERNAME)%'
+                password: '%env(REPLICAS_PASSWORD)%'
+            toolforge_s7:
+                host: '%env(REPLICAS_HOST_S7)%'
+                port: '%env(REPLICAS_PORT_S7)%'
+                user: '%env(REPLICAS_USERNAME)%'
+                password: '%env(REPLICAS_PASSWORD)%'
+            toolforge_s8:
+                host: '%env(REPLICAS_HOST_S8)%'
+                port: '%env(REPLICAS_PORT_S8)%'
+                user: '%env(REPLICAS_USERNAME)%'
+                password: '%env(REPLICAS_PASSWORD)%'
     orm:
         auto_generate_proxy_classes: '%kernel.debug%'
         default_entity_manager: default
--- a/src/BookProvider.php
+++ b/src/BookProvider.php
@@ -2,11 +2,11 @@
 
 namespace App;
 
+use App\Repository\CreditRepository;
 use App\Util\Api;
 use App\Util\Util;
 use DOMDocument;
 use GuzzleHttp\Pool;
-use GuzzleHttp\Promise\PromiseInterface;
 use GuzzleHttp\Psr7\Response;
 use Iterator;
 
@@ -18,15 +18,16 @@ class BookProvider {
 	protected $options = [
 		'images' => true, 'fonts' => false, 'categories' => true, 'credits' => true
 	];
-	private $creditUrl = 'https://phetools.toolforge.org/credits.py';
+	private $creditRepo;
 
 	/**
 	 * @param $api Api
 	 * @param bool[] $options
 	 */
-	public function __construct( Api $api, array $options ) {
+	public function __construct( Api $api, array $options, CreditRepository $creditRepo ) {
 		$this->api = $api;
 		$this->options = array_merge( $this->options, $options );
+		$this->creditRepo = $creditRepo;
 	}
 
 	/**
@@ -114,7 +115,9 @@ class BookProvider {
 			$book->categories = $this->getCategories( $metadataSrc );
 		}
 		$pageTitles = $parser->getPagesList();
-		$namespaces = $this->api->getNamespaces();
+		$namespacesInfo = $this->api->getNamespaces();
+		$namespaces = array_keys( $namespacesInfo );
+
 		if ( !$isMetadata ) {
 			if ( !$parser->metadataIsSet( 'ws-noinclude' ) ) {
 				$book->content = $parser->getContent( true );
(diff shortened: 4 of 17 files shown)


src/Repository/CreditRepository.php
```diff
@@ -0,0 +1,73 @@
...
+	public function getPageCredits( ReplicasClient $replicasClient, $lang, $namespaces, $pages ) : array {
+		$dbname = $this->getDatabaseName( $lang, 'wikisource' );
+
+		$conn = $replicasClient->getConnection( $dbname );
+		$pageNames = [];
+		$namespaceName = "";
+		$splittedPages = [];
+	
+		foreach ( $pages as $page ) {
+			$splitPage = explode( ":", $page, 2 );
```
MusikAnimal, line 32: What happens if I have a page called `Foo:Baz`? It looks like it will assume `Foo` is a valid namespace when it is not. The code below should somehow account for this, I think.

src/Repository/CreditRepository.php
```diff
@@ -0,0 +1,73 @@
...
+			if ( isset( $splitPage[1] ) ) {
+				$namespaceName = $splitPage[0];
+				$pageNames[] = $splitPage[1];
+				$namespaceId = $namespaces[$namespaceName];
+				$splittedPages[$namespaceId][] = "'" . $splitPage[1] . "'";
```
MusikAnimal, line 37: This will presumably cause an SQL error for pages with an apostrophe in the title. I usually use Doctrine's `Connection::quote()` method for this, so something like:
```
$splittedPages[$namespaceId][] = $conn->quote($splitPage[1])
```

src/Repository/CreditRepository.php
```diff
@@ -0,0 +1,73 @@
...
+			SELECT 1
+			FROM enwikisource_p.user_groups
+			WHERE ug_user = user_id
+				AND ug_group = 'bot'
+			) AS bot
+			FROM
+			enwikisource_p.revision_userindex
+			JOIN enwikisource_p.page ON page_id = rev_page
+			JOIN enwikisource_p.actor_revision ON actor_id = rev_actor
+			JOIN enwikisource_p.user ON user_id = actor_user
```
MusikAnimal, line 64: the `enwikisource_p.` prefix should be removed, since the database is set when you call `$replicasClient->getConnection( $dbname );`

src/Repository/CreditRepository.php
```diff
@@ -0,0 +1,73 @@
...
+		if ( $family === 'wikisource' ) {
+			if ( in_array( $domain, [ 'old', 'www', 'mul' ] ) ) {
+				$dbname = 'sourceswiki_p';
+			} elseif ( $domain === 'zh-min-nan' ) {
+				$dbname = 'zh_min_nanwikisource_p';
+			} else {
+				$dbname = $domain . $family . '_p';
+			}
+		} else {
+			$dbname = $domain . $family . '_p';
```
MusikAnimal, line 18: A note that the `_p` isn't strictly needed here since ReplicasClient will do that for you, but you do otherwise normally need to append `_p`.

src/Repository/CreditRepository.php
```diff
@@ -0,0 +1,73 @@
...
+		}
+
+		$whereSql = [];
+
+		foreach ( $splittedPages as $namespaceId => $pages ) {
+			$whereSql[] = "( page_namespace = $namespaceId AND page_title IN (" . implode( ",", $pages ) . ")" . ")";
+		}
+		$whereSqlString = "(" . implode( ' OR ', $whereSql ) . ")";
+
+		$stmt = $conn->prepare( "SELECT
```
MusikAnimal, line 51: At this point we're not actually preparing anything, so you could use `$conn->executeQuery( ...sql... )->fetchAll()`. Typically you do want to use prepared statements to guard against SQL injection, but I think Doctrine's `Connection::quote()` method will do that work, as suggested above on line 37

MusikAnimal (commented):
The approach looks pretty good to me! I know you're not done with this yet but I still left a few comments.

.env.test
```diff
@@ -1 +1,2 @@
+DATABASE_URL=mysql://root:@database:3306/wsexport_test
```
MusikAnimal, line 1: I _think_ this only applies to Docker, which not everyone uses. Maybe it should go in `.env.test.local` instead, assuming that's a thing?

src/Command/ExportCommand.php
```diff
@@ -19,15 +21,23 @@ class ExportCommand extends Command {
...
+	private $creditRepo;
+
 	/** @var Api */
 	private $api;
 
 	/** @var bool */
 	private $enableCache;
 
-	public function __construct( GeneratorSelector $generatorSelector, Api $api, bool $enableCache ) {
+	public function __construct( GeneratorSelector $generatorSelector, ReplicasClient $replicasClient, CreditRepository $creditRepo, Api $api, bool $enableCache ) {
```
MusikAnimal, line 36: Not critical, but another option may be to only inject `ReplicasClient` in the `CreditRepository` constructor, since that's the only place it's used (if I'm reading the code correctly). That way you don't need to pass around the `ReplicasClient` instance.

So CreditRepository could have:
```
/** @var ReplicasClient $client */
private $client;

public function __construct( ReplicasClient $client ) {
	$this->client = $client;
}
```
That may or may not work out-of-the-box, but if not, you can specify this dependency in services.yaml like:

```
services:
	...
	App\Repository\CreditRepository:
		arguments:
			- '@Wikimedia\ToolforgeBundle\Service\ReplicasClient'
```

src/Repository/CreditRepository.php
MusikAnimal: Just a note that if you need to query a different database, such as `commonswiki`, you'll need to get a separate Connection instance as with `$replicasClient->getConnection( 'commonswiki' )`

src/Repository/CreditRepository.php
cimurah: Good point! Done.

src/Repository/CreditRepository.php
cimurah: Done. Thank you!

src/Repository/CreditRepository.php
cimurah: Done.

src/Repository/CreditRepository.php
cimurah: Done!

.env.test
cimurah: Yup! Done

src/Command/ExportCommand.php
cimurah: I had tried your first suggestion, and for whatever reason I couldn't get it to work. I tried it again and it worked! :)

config/packages/doctrine.yaml
```diff
@@ -5,6 +5,46 @@ doctrine:
                 url: '%env(resolve:DATABASE_URL)%'
                 driver: pdo_mysql
                 charset: utf8mb4
+            toolforge_s1:
+                host: '%env(REPLICAS_HOST)%'
```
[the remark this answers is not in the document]
cimurah: Done. Thank you!

src/BookProvider.php
```diff
@@ -159,7 +163,7 @@ public function getMetadata( string $title, bool $isMetadata, DOMDocument $doc )
 			$book->chapters = $chapters;
 
 			if ( $this->options['credits'] ) {
-				$creditPromises = $this->startCredits( $book, $chapterTitles, $pageTitles, $pictures );
+				$creditPromises = $this->startCredits( $book, $namespacesInfo, $chapterTitles, $pageTitles, $pictures );
```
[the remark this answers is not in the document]
cimurah: Good idea! Done

src/BookProvider.php
```diff
@@ -321,66 +326,46 @@ protected function startCredits( Book $book, array $chapters, array $otherPages,
...
+		return array_merge( $pageCredits, $imageCredits );
 	}
 
 	/**
-	 * @param PromiseInterface[] $promises
+	 * @param array $creditResults
 	 * @return array
 	 */
-	public function finishCredit( $promises ) {
+	public function finishCredit( $creditResults ) {
```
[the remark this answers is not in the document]
cimurah: Great point! I removed this `finishCredit` method but I'm afraid the results still need to be ordered since the image credits and page credits might return the same `actor names` so we need to combine and count their contributions? Does this make sense?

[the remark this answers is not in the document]
samwilson: Oh yes, I totally wasn't thinking correctly. Good point. I wonder if the two queries could be done in one request, with a UNION. Then the sorting can be done by the DB. (Not sure it matters too much though.)

.env
```diff
@@ -41,3 +41,17 @@ APP_LOG_RECIPIENT_1=admin1@example.org
 APP_LOG_RECIPIENT_2=admin2@example.org
 APP_LOG_SUBJECT="[Wikisource Export]"
 EPUBCHECK_JAR="/usr/bin/epubcheck"
+# For production, REPLICAS_HOST should be *.web.db.svc.eqiad.wmflabs (or analytics instead of web)
+# For local environments, use 127.0.0.1 and host.docker.internal for docker.
+REPLICAS_HOST="127.0.0.1"
```
MusikAnimal, line 46: Sorry, I got this wrong. In production we'll need a separate env variable for each host (s1 through s8). See my update to the ToolforgeBundle README https://github.com/wikimedia/ToolforgeBundle/pull/46

So here it should be:
```
REPLICAS_HOST_S1="127.0.0.1"
REPLICAS_HOST_S2="127.0.0.1"
REPLICAS_HOST_S3="127.0.0.1"
REPLICAS_HOST_S4="127.0.0.1"
REPLICAS_HOST_S5="127.0.0.1"
REPLICAS_HOST_S6="127.0.0.1"
REPLICAS_HOST_S7="127.0.0.1"
REPLICAS_HOST_S8="127.0.0.1"
```
also change doctrine.yaml accordingly.

src/Repository/CreditRepository.php
```diff
@@ -0,0 +1,131 @@
...
+		}
+		return $dbname;
+	}
+
+	public function getPageCredits( $lang, $namespaces, $pages ) : array {
+		$dbname = $this->getDatabaseName( $lang, 'wikisource' );
+
+		$connection = $this->client->getConnection( $dbname );
+		$pageNames = [];
+		$namespaceName = "";
```
MusikAnimal, line 34: Unused variable

.env
cimurah: You didn't get it wrong. You figured this out before we did :) Thank you!

src/Repository/CreditRepository.php
```diff
@@ -0,0 +1,131 @@
...
+		$splittedPages = [];
```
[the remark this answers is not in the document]
cimurah: Great suggestion! I started naming the variables like they did in the python version, but i'll probably get confused too when we look back in a week. Done.

src/Repository/CreditRepository.php
```diff
@@ -0,0 +1,131 @@
...
+
+class CreditRepository {
+	/** @var ReplicasClient */
+	private $client;
+
+	public function __construct( ReplicasClient $client ) {
+		$this->client = $client;
+	}
+
+	public function getDatabaseName( $domain, $family ) {
```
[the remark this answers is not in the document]
cimurah: Yup! Done.

src/Repository/CreditRepository.php
```diff
@@ -0,0 +1,131 @@
...
+		$pageNames = [];
+		$namespaceName = "";
+		$splittedPages = [];
+
+		foreach ( $pages as $page ) {
+			$splitPage = explode( ":", $page, 2 );
+			if ( isset( $splitPage[1] ) ) {
+				$namespaceName = $splitPage[0];
+				$pageNames[] = $splitPage[1];
+				if ( isset( $namespaces[$namespaceName] ) ) {
```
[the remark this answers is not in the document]
cimurah: Agree. I followed your logic and it's correct. I just have a minor question. For the cases like `Foo:Bar`, do we want to assume that the page might be in the main namespace or should we omit it completely since `Foo` is not a valid namespace?

src/Repository/CreditRepository.php
cimurah: Removed variable. Thank you!

src/BookProvider.php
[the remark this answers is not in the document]
cimurah: I left the queries separately  so we don't have a huge query. But do think is better to make one request? Do you think that might be a better approach for performance maybe?

samwilson (changes requested):
The `app:check` command needs to be updated with the $creditsRepo too. (My fault for not including a test for it when I added the command!)

src/Repository/CreditRepository.php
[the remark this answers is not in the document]
MusikAnimal: Any page can have a colon in the title, so no, they should never be skipped. A real-worl example is [John Huss: his life, teachings and death, after five hundred years](https://en.wikisource.org/wiki/John_Huss:_his_life,_teachings_and_death,_after_five_hundred_years). Obviously `John Huss` is not a namespace, hence the full page title (before and after the colon) is a mainspace title.

src/BookProvider.php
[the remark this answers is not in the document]
MusikAnimal: You can't! The image query is on commonswiki (s7) while the wikisource databases live on a different db section. In my testing both queries are pretty fast (at least on the new cluster), so combining them even if we could probably wouldn't see significant gains.

src/BookProvider.php
```diff
@@ -19,14 +19,16 @@ class BookProvider {
 		'images' => true, 'fonts' => false, 'categories' => true, 'credits' => true
 	];
 	private $creditUrl = 'https://phetools.toolforge.org/credits.py';
```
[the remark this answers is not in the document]
cimurah: Done. Thank you :)

dayllanmaza:
Could we update  README.md with instructions on how to setup the replica's connection? At the very least we should have a link to the ToolforgeBundle docs

src/Repository/CreditRepository.php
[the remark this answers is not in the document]
cimurah: Got it. I did not know that a page title's can contain a color. Thanks!

cimurah:
> The `app:check` command needs to be updated with the $creditsRepo too. (My fault for not including a test for it when I added the command!)

Done :)

README.md
```diff
@@ -43,6 +43,19 @@ Installation
...
+  * Establish an SSH tunnel to the replicas (only necessary on local environments)
+```bash
+$ ./php bin/console toolforge:ssh
+```
+  * Bind address for docker enviroments
+```bash
+$ php bin/console toolforge:ssh --bind-address=0.0.0.0
+```
+
+7. Set replicas credentials in the `.env.local` file.
```
samwilson, line 57: This needs to happen before running the `toolforge:ssh` command.

cimurah:
> Could we update README.md with instructions on how to setup the replica's connection? At the very least we should have a link to the ToolforgeBundle docs

Good point! I added instructions regarding toolforge bundle and a link to the multiple connections documentation.

README.md
cimurah: I revised the documentation. Let me know if I need to make any other changes. Thank you!

PR #336, opened 2021-02-24
Change api cache ttl to be a .evn var

dayllanmaza:


samwilson:
Looks like Refresh needs to be updated too.

PR #342, opened 2021-03-12
Create option to disable credits in export form

cimurah:
Add option that will exclude/include credits in exported books.

Bug: T274959

### DIFF
--- a/src/Generator/EpubGenerator.php
+++ b/src/Generator/EpubGenerator.php
@@ -9,6 +9,7 @@ use App\Util\Api;
 use App\Util\Util;
 use Exception;
 use IntlDateFormatter;
+use Krinkle\Intuition\Intuition;
 use ZipArchive;
 
 /**
@@ -33,9 +34,13 @@ class EpubGenerator implements FormatGenerator {
 	/** @var Api */
 	protected $api;
 
-	public function __construct( FontProvider $fontProvider, Api $api ) {
+	/** @var Intuition */
+	private $intuition;
+
+	public function __construct( FontProvider $fontProvider, Api $api, Intuition $intuition ) {
 		$this->fontProvider = $fontProvider;
 		$this->api = $api;
+		$this->intuition = $intuition;
 	}
 
 	/**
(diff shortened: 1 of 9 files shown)


samwilson (changes requested):
Looks good to me, works well.

src/Generator/EpubGenerator.php
```diff
@@ -398,17 +403,23 @@ private function getXhtmlTitle( Book $book ) {
...
 
 	private function getXhtmlAbout( Book $book, $wsUrl ) {
-		$list = '<ul>';
-		$listBot = '<ul>';
-		foreach ( $book->credits as $name => $value ) {
-			if ( $value[ 'bot' ] !== null ) {
-				$listBot .= '<li>' . htmlspecialchars( $name, ENT_QUOTES ) . "</li>\n";
-			} else {
-				$list .= '<li>' . htmlspecialchars( $name, ENT_QUOTES ) . "</li>\n";
+		if ( empty( $book->credits ) ) {
```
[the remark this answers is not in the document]
cimurah: Good catch! Done.

src/Generator/EpubGenerator.php
```diff
@@ -398,17 +403,23 @@ private function getXhtmlTitle( Book $book ) {
...
-		$list = '<ul>';
-		$listBot = '<ul>';
-		foreach ( $book->credits as $name => $value ) {
-			if ( $value[ 'bot' ] !== null ) {
-				$listBot .= '<li>' . htmlspecialchars( $name, ENT_QUOTES ) . "</li>\n";
-			} else {
-				$list .= '<li>' . htmlspecialchars( $name, ENT_QUOTES ) . "</li>\n";
+		if ( !$book->options['credits'] ) {
+			$list = $this->intuition->msg( 'credits-default-message' ) ??
+					'credits-default-message missing';
```
samwilson, line 408: Why do we need to fall back to this?

src/Generator/EpubGenerator.php
```diff
@@ -398,17 +403,23 @@ private function getXhtmlTitle( Book $book ) {
...
 	private function getXhtmlAbout( Book $book, $wsUrl ) {
-		$list = '<ul>';
-		$listBot = '<ul>';
-		foreach ( $book->credits as $name => $value ) {
-			if ( $value[ 'bot' ] !== null ) {
-				$listBot .= '<li>' . htmlspecialchars( $name, ENT_QUOTES ) . "</li>\n";
-			} else {
-				$list .= '<li>' . htmlspecialchars( $name, ENT_QUOTES ) . "</li>\n";
+		if ( !$book->options['credits'] ) {
+			$list = $this->intuition->msg( 'credits-default-message' ) ??
```
samwilson, line 407: The language of Intuition should be set before this, so that it uses the book's language rather than the interface language.

samwilson (commented):
It's probably not that common to have the interface language differ from the book language, but it's not hard to support so we might as well.

src/Generator/EpubGenerator.php
cimurah: Phan will complain since the `msg` function might return null, and null cannot be passed to `Util::getXhtmlFromContent` as argument 2. I added a comment in the code so there is not confusion in the future.

src/Generator/EpubGenerator.php
cimurah: Good point! Done.

samwilson (approved):

PR #343, opened 2021-03-15
Improve errors' UI

samwilson:
* Add new WsExportException that can be used to throw
  user-recoverable translatable errors.
* Fix various error-handling to use this new system.
* Move all HTTP-related errors out of the main export classes,
  and keep all that in the Controllers.
* Move CSS into its own file.

Bug: https://phabricator.wikimedia.org/T277237


src/Controller/ExportController.php
```diff
@@ -99,8 +95,16 @@ public function home(
...
+		$exception = false;
+		$response = new Response();
 		if ( $request->get( 'page' ) ) {
-			return $this->export( $request, $api, $fontProvider, $generatorSelector, $creditRepo );
+			try {
+				return $this->export( $request, $api, $fontProvider, $generatorSelector, $creditRepo );
+			} catch ( WsExportException $ex ) {
+				$exception = $ex;
+				// @TODO Return different status codes for different errors (e.g. timeout vs not found).
+				$response->setStatusCode( 404 );
```
MusikAnimal, line 106: If we're going to use a generic response code I would use a 400 or 500. A 404 should strictly be for not found, I think...

If it helps, maybe `WsExportException` could take an optional third argument that sets the status code that should returned here?

src/OpdsBuilder.php
```diff
@@ -55,7 +60,7 @@ public function buildFromCategory( $categoryTitle ) {
 			'gcmlimit' => '100',
 		] );
 		if ( !array_key_exists( 'query', $response ) ) {
-			throw new NotFoundHttpException( "Category not found: $categoryTitle" );
+			throw new Exception( "Category not found: $categoryTitle" );
```
MusikAnimal, line 63: Did we want this to be translatable, too? I think it's okay if we don't (probably a rare error), but as written the changes to this file don't have any user-facing effect, do they?

public/styles.css
```diff
@@ -0,0 +1,39 @@
...
+}
+
+.error > .message {
+	flex: 20;
+	padding-top: 0.4rem;
+}
+
+.error p {
+	margin: 0;
+}
```
MusikAnimal, line 39: Nitpick, maybe also add
```
.error glyphicon {
    top: 9px;
}
```
to make the (X) icon in the error message more vertically aligned with the text. Tested in Chrome and Firefox.

MusikAnimal (commented):
Looking good! Just a few minor concerns, mainly with using 404 as a generic response code. My feedback should be taken only as an opinion, though :)

src/WsExportException.php
```diff
@@ -0,0 +1,29 @@
...
+
+	public function getI18nMessage(): string {
+		return 'exception-' . $this->i18nMessage;
+	}
+
+	public function getI18nParams(): array {
+		return $this->i18nParams;
+	}
+
+}
```
MusikAnimal, line 29: Should normally go in a directory called Exception. I know there's only one right now, but we might add more (I was thinking we need a wrapper for `ProcessFailedException` to keep the error message title consistent so the emails are grouped)

public/styles.css
samwilson: Good point. Actually, if we `align-items: center;` then it's always in the right place (I think).

src/Controller/ExportController.php
samwilson: Yeah I was thinking that would be good; was going to ask about it. Will change it now.

src/OpdsBuilder.php
samwilson: That's right: this is only used in a Command, and none of those are translated.

src/WsExportException.php
samwilson: Done.

templates/bundles/TwigBundle/Exception/error.html.twig
```diff
@@ -0,0 +1,32 @@
...
+            </p>
+            <p>
+                <strong>{{ msg('error-page-details', [exception.message]) }}</strong>
+            </p>
+            <p>
+                {% set issue_link %}{#
+                    #}<a href="https://phabricator.wikimedia.org/maniphest/task/edit/form/1/?projects=ws-export">{#
+                        #}{{ msg( 'error-page-issue-link' ) }}{#
+                    #}</a>{#
+                #}{% endset %}
```
MusikAnimal, line 26: Not sure if it works here but I think you can [remove the whitespace](https://symfony.com/blog/better-white-space-control-in-twig-templates) with:
```
{%- set issue_link %}
    <a href="https://phabricator.wikimedia.org/maniphest/task/edit/form/1/?projects=ws-export">
        {{- msg( 'error-page-issue-link' ) -}}
    </a>
{% endset -%}
```

samwilson: Oh, that's a cool trick! Thanks. Done. It's only required at the end, to stop the period floating out on its own.

PR #349, opened 2021-03-19
Tweak UI (layout, logo, text styles)

samwilson:


public/img/Wikisource-logo.svg
```diff
@@ -0,0 +1,30 @@
...
+    <path id="path2468" d="M 288.03125,202.3125 C 281.27622,204.13028 273.10387,211.41115 267.5625,224.9375 C 265.0321,230.27457 262.54791,233.27287 260.1875,234.65625 C 260.42406,257.90517 265.94713,274.45688 277.375,283.53125 C 273.9859,296.98175 270.60724,315.52359 276.75,326.75 C 266.37085,323.46682 270.60263,344.3345 267.53125,353.125 C 262.34167,367.74053 257.77809,367.10409 252.90625,368.375 C 257.35445,373.45867 263.07757,378.85396 271.65625,382.03125 C 271.65774,382.04785 271.65317,382.07711 271.65625,382.09375 C 271.67362,382.16873 271.73429,382.26787 271.78125,382.34375 C 272.42051,383.3146 275.29796,384.37175 276.6875,385.46875 C 279.16461,386.15238 282.58538,385.54582 285.0625,381.78125 C 288.48532,374.33338 287.91001,368.15617 297.125,361.34375 C 302.46044,357.49908 308.74119,343.00137 309.21875,342.28125 C 290.98447,332.03327 282.36549,312.5039 285.3125,293.6875 C 286.15977,288.49794 289.55411,282.67936 285,274.3125 C 279.55423,251.31921 283.02238,224.56744 288.03125,202.3125 z " style="fill:#236798;fill-opacity:1;fill-rule:evenodd;stroke:none;stroke-width:1px;stroke-linecap:butt;stroke-linejoin:miter;stroke-opacity:1"/>
+  </g>
+</svg>
```
MusikAnimal, line 30: This looks like it could benefit from some compression. I ran it through https://vecta.io/nano and it shaved off 8.6 KB (less than half of the original size).

MusikAnimal (commented):
Looks good! I left a recommendation about compressing the SVG.

Also, it looks like we didn't restrict the width of the form as seen in the [mockup](https://phabricator.wikimedia.org/F34171015) – though I worry there might be some word wrapping for the more verbose languages like German. Also it might look odd to have the form flushed to the left rather than centered?

public/img/Wikisource-logo.svg
samwilson: I thought about doing that, but there's some history on Commons about how you're not supposed to modify these. However, that's from 2012 and I've now made a note in the README so I think we're fine.

samwilson:
> Also, it looks like we didn't restrict the width of the form as seen in the mockup – though I worry there might be some word wrapping for the more verbose languages like German. Also it might look odd to have the form flushed to the left rather than centered?

No, I'm, doing that in a subsequent patch (along with a few other things; I left a note on the phab ticket).

PR #351, opened 2021-03-22
Add urldecode to images and pages names in credit repo

cimurah:
Add urldecode to names in credit repository queries so that it decodes
any urlencoded characters.

Bug: T257886


MusikAnimal (approved):
Works in my testing!

PR #352, opened 2021-03-23
Further improve handling of common errors

MusikAnimal:
This covers errors thrown when doing format conversions, MySQL timeout
and max_user_connections errors.

This also hides the 'Learn more' link for 404 errors since re-trying
won't help in that case.

Bug: https://phabricator.wikimedia.org/T278189

### DIFF
--- a/src/Generator/ConvertGenerator.php
+++ b/src/Generator/ConvertGenerator.php
@@ -3,11 +3,14 @@
 namespace App\Generator;
 
 use App\Book;
+use App\Exception\WsExportException;
 use App\FontProvider;
 use App\Util\Api;
 use App\Util\Util;
 use InvalidArgumentException;
 use Krinkle\Intuition\Intuition;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Process\Exception\ProcessTimedOutException;
 use Symfony\Component\Process\Process;
 
 /**
@@ -128,7 +131,7 @@ class ConvertGenerator implements FormatGenerator {
 
 	/**
 	 * create the file
-	 * @param $data Book the title of the main page of the book in Wikisource
+	 * @param $book
 	 * @return string
 	 */
 	public function create( Book $book ) {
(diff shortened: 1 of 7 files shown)


MusikAnimal:
This doesn't have logging, but I question if we really need to log these issues. Maybe the `max_user_connections` error; because if that happens a LOT, we can request an increase in quota. The other things we're capturing aren't indicative of bugs or something we can solve, rather they are intermittent issues where re-trying is likely to help, so I think it's okay to suppress emails and refrain from logging.

The task https://phabricator.wikimedia.org/T278189 talks about capturing _all_ 500-level errors, but I'm not convinced that's a good idea. We can simulate that the error was capturing by basically moving the stuff inside the `{% if exception %}` block from export.html.twig to error.html.twig, that way it always _looks_ like it's captured. But that's sort of undoing a lot of the work @samwilson did.

src/Generator/ConvertGenerator.php
```diff
@@ -154,13 +157,17 @@ private function createEpub( Book $book ) {
...
-		$process->mustRun();
+		try {
+			$command = array_merge(
+				[ $this->getEbookConvertCommand(), $epubFileName, $outputFileName ],
+				explode( ' ', self::$CONFIG[$this->format]['parameters'] )
+			);
+			$process = new Process( $command );
+			$process->setTimeout( $this->timeout ?? 120 );
+			$process->mustRun();
+		} catch ( RuntimeException $e ) {
```
samwilson, line 168: Do we want to catch *all* RuntimeExceptions? I thought we'd be catching ProcessTimedOutException here (as something that's common, and not something we can do much about). Other ones, such as ProcessFailedException, we might still want to treat as 500s, because it could be due to e.g. the disk filling up or something.

src/Repository/CreditRepository.php
```diff
@@ -142,8 +143,31 @@ public function getImageCredits( array $images ) : array {
...
+	}
+
+	/**
+	 * Runs Connection::fetchAllAssociative() on the given SQL and closes the Connection.
+	 * Also captures common MySQL errors where we want to tell the user to re-try.
+	 * @param Connection $connection
+	 * @param string $sql
+	 * @return array
+	 * @throws WsExportException
+	 * @throws \Doctrine\DBAL\Exception
```
samwilson, line 155: I think we're only adding directly-thrown exceptions with `@throws` (i.e. WsExportException in this case, but not this other one), or at least that's what I've been doing!

samwilson (commented):
I think this looks good, and is pretty much what I had in mind with the work already done with exception handling. I think it's fine to not log/email the errors that we know about and can't do much about.

I'd thought about adding something to the WsExportException to make it possible to alter the 'learn more' list to make it more appropriate for the error — e.g. for conversion timeouts highlight the EPUB option, but don't show that if they're already asking for an EPUB. That can be for later though, if we want it at all.

src/Generator/ConvertGenerator.php
MusikAnimal: Good call! I've changed it to just capture just `ProcessTimedOutException`.

PR #353, opened 2021-03-25
Fix layout of 'back to Wikisource' button

samwilson:
Also add missing message.

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -18,6 +18,7 @@
 	"issues": "Please report issues under the $1 project on Phabricator.",
 	"issue-button": "Create issue",
 	"back-to-wikisource": "Back to Wikisource",
+	"previous-alt-text": "Backwards-facing arrow icon that indicates that the link returns to Wikisource.",
 	"export-book": "Export a book",
 	"lang-field-label": "Wikisource",
 	"lang-field-value": "$1 ($2)",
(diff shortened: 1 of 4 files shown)


MusikAnimal (commented):
As I understood it, we wanted to replace the space before the text with padding, such that there isn't an underline when hovering over it. I think what we have here is hugging the arrow icon too much, but my understanding of what Nicolas said could be wrong!

The alignment meanwhile looks perfect in Chrome but seems even more off now in Firefox, at on my end:
Screenshot from 2021-03-24 22-10-57

samwilson:
> As I understood it, we wanted to replace the space before the text with padding, such that there isn't an underline when hovering over it. I think what we have here is hugging the arrow icon too much, but my understanding of what Nicolas said could be wrong!
> 
> The alignment meanwhile looks perfect in Chrome but seems even more off now in Firefox, at on my end:
> Screenshot from 2021-03-24 22-10-57

Oh, I totally thought it was about *reducing* the gap between them. But it makes sense, what you say! I'll update it to increase the space, and move the text down a bit also (although I'm always worried about pixel-perfect moving of text; it never seems to work everywhere!).

I did wonder about the hover underline. @nayoub does that stay or go?

nayoub:
Awesome! Yes absolutely @samwilson 
The hover underline stays, the only issue on this was that we had an empty space underlined on hover before the 'B' on "Back to Wikisource".

samwilson:
Okay cool, thanks. I've moved the underline to just be on the text bit.

PR #356, opened 2021-03-29
Sanitize cache keys

MusikAnimal:
This prevents InvalidArgumentExceptions that happen from erroneous user
input for the lang= parameter.

### DIFF
--- a/src/Util/Util.php
+++ b/src/Util/Util.php
@@ -230,4 +230,13 @@ class Util {
 		$document->encoding = 'UTF-8';
 		return $document;
 	}
+
+	/**
+	 * Remove reserved characters from a cache key.
+	 * @param string $key
+	 * @return string
+	 */
+	public static function sanitizeCacheKey( string $key ): string {
+		return preg_replace( '/[{}()\/\@\:"]/', '-', $key );
+	}
 }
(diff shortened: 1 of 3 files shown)


MusikAnimal:
Another option of course is to validate the lang= parameter, but if we ever build cache keys from something other than just the language, you'll want to do this same sanitizing. Happy to go with `lang` validation for now if people think that is a better solution.

src/Util/Util.php
```diff
@@ -230,4 +230,13 @@ public static function buildDOMDocumentFromHtml( string $html ): DOMDocument {
...
 		return $document;
 	}
+
+	/**
+	 * Remove reserved characters from a cache key.
+	 * @param string $key
+	 * @return string
+	 */
+	public static function sanitizeCacheKey( string $key ): string {
+		return preg_replace( '/[{}()\/\@\:"]/', '', $key );
```
samwilson, line 240: Should probably replace with a hyphen or something, to avoid a completely empty key. (As I did somewhere recently, much annoyingly.)

MusikAnimal: Done

samwilson (approved):
I guess lang validation would be tricky, because we want to cache the valid-langs list! But anyway, we can follow up with that; this is still useful and solves the current issue. :)

PR #359, opened 2021-03-31
Reduce width of form and fix label font sizes

samwilson:

### DIFF
--- a/public/styles.css
+++ b/public/styles.css
@@ -3,12 +3,15 @@ header {
 	margin-bottom: 25px;
 }
 
+.container {
+	max-width: 585px;
+}
+
 header .container {
 	display: flex;
 	align-items: center;
-	max-width: 100%;
-	/* 1170 is bootstrap's widest breakpoint, and the 85 and 20 are for the logo. */
-	width: calc( 1170px + ( ( 85px + 20px ) * 2 ) );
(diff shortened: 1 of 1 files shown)


public/styles.css
```diff
@@ -3,12 +3,16 @@ header {
...
+}
+
 header .container {
 	display: flex;
 	align-items: center;
 	max-width: 100%;
-	/* 1170 is bootstrap's widest breakpoint, and the 85 and 20 are for the logo. */
-	width: calc( 1170px + ( ( 85px + 20px ) * 2 ) );
+	/* 585 is the container width set above, and the 85 and 20 are for the logo. */
+	width: calc( 585px + ( ( 85px + 20px ) * 2 ) );
```
MusikAnimal, line 15: This an above should maybe be `max-width` to retain the responsive design. Otherwise content exceeds the viewport on mobile devices.

MusikAnimal (changes requested):
Seems to meet the acceptance criteria! I'm assuming @nayoub wanted the content to be centered (this is unclear from the mockup)?

Screenshot from 2021-03-30 22-38-05

The only issue I see is this breaks the mobile responsiveness, which I think can be fixed by simply using `max-width` instead of `width`.

nayoub:
Yes exactly, centered looks great! Thanks @MusikAnimal & @samwilson!

PR #360, opened 2021-04-07
Improve subpage-finding to not use URLs

samwilson:
This changes the chapter subpage finding logic (when there's no
ws-summary in use) to use the links' title attribute rather than
their href attribute. It also moves the prefix-searching out of
XPath and into the link loop.

Bug: https://phabricator.wikimedia.org/T275870

### DIFF
--- a/src/PageParser.php
+++ b/src/PageParser.php
@@ -5,7 +5,6 @@ namespace App;
 use DOMDocument;
 use DOMElement;
 use DOMXPath;
-use Wikimedia\ToolforgeBundle\Service\Util;
 
 class PageParser {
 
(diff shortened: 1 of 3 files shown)


samwilson:
It might be that we can combine the two chapter-finding methods of PageParser, but I didn't try to tackle that just yet. Happy to if anyone thinks it's worth it.

src/PageParser.php
```diff
@@ -108,32 +107,37 @@ public function getChaptersList( $pageList, $namespaces ) {
...
-			$list = $this->xPath->query( '//a[contains(@href,"/' . Util::wfUrlencode( $title ) . '/") and
+			// If no chapters were found in the usual way, try to guess the chapters
+			// by finding all links that start with the current page's name.
+			$list = $this->xPath->query( '//a[
 				not(
 					contains(@class,"new") or
 					contains(@class,"extiw") or
 					@rel = "mw:WikiLink/Interwiki" or
 					contains(@class,"external") or
-					contains(@href,"#") or
```
Tpt, line 123: it would be nice to have tests to check what happen with fragment links to the same page and to a subpage. I don't remind what is the Parsoid output for them. ws-export had troubles with them in the past.

samwilson: Good idea. Done. It doesn't clash because in this patch I'm adding the prefix check with the trailing slash, which will never happen for a self-link.

Tpt (approved):
Thank you!

samwilson:
:-)

PR #361, opened 2021-04-07
Remove `mw:Extension` link elements

samwilson:
These are not valid in EPUBs.

Bug: https://phabricator.wikimedia.org/T272819


Tpt (approved):
What about removing all <meta> and <link>? I am not sure there is any proper usecase for them.

samwilson:
Hmm that's a good point. These are the only ones that epubcheck is complaining about (although, maybe this *is* all of them, I've not really looked deeply yet...)

PR #369, opened 2021-06-17
Remove $wsexportConfig global and inject FileCache

samwilson:
This makes FileCache injectable and switches it to using the
kernel's $projectDir to figure out where to store files, rather
than the $wsexportConfig['tempPath'] value.

Bug: https://phabricator.wikimedia.org/T265854


MusikAnimal (approved):
Looks good and works great in my testing!

PR #372, opened 2021-07-28
Limit the size of book covers

inductiveload:
Currently, non-Djvu/Pdf covers use the full size of the image
specified. This can be very large, much larger than the book
itself.

This commit instead limits the cover size to 400px (or the image
size, if smaller).

It also avoids hardcoding assumptions about the thumbnail URL
and instead uses the 'thumburl' field.

It also adds better defence against bad page numbers (and defaults
to 1 in that case), as well as defence against a non-JPG thumbnail
for a DjVu/PDF file.

This does mean there are two requests to get the cover, as it's
not possible to tell the image mime type and max size and get
the appropiate thumbnail in a single request.

Bug: T286389

### DIFF
--- a/src/BookProvider.php
+++ b/src/BookProvider.php
@@ -275,30 +275,49 @@ class BookProvider {
 		$title = $id[0];
 		$picture = new Picture();
 		$picture->title = $cover;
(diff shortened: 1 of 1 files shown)


samwilson (approved):
Looks great! Sorry for being so slow to review this.

I've tested this on a few dozen random books, and all seems groovy.

PR #373, opened 2021-07-28
Avoid blindly re-encoding HTML files

inductiveload:
Previously, HTML files werei stripped of their XML
Processing Instruction headers and re-encoded from
UTF-8 to HTML-ENTITIIES to be fed into the DomDocument.

This caused problems for documents with CDATA blocks that
contained Unicode, as it's not correct to escape that as
HTML entities in the general case. For example CSS or
binary data doesn't use that escaping system.

Instead, load it directly and then remove the PI nodes
after the fact.

Bug: https://phabricator.wikimedia.org/T271390

### DIFF
--- a/src/Util/Util.php
+++ b/src/Util/Util.php
@@ -185,10 +185,26 @@ class Util {
 	 */
 	public static function buildDOMDocumentFromHtml( string $html ): DOMDocument {
 		$document = new DOMDocument( '1.0', 'UTF-8' );
+
 		libxml_use_internal_errors( true );
-		$document->loadHTML( mb_convert_encoding( str_replace( '<?xml version="1.0" encoding="UTF-8" ?>', '', $html ), 'HTML-ENTITIES', 'UTF-8' ) );
+		// enforce a UTF-8 encoding declaration
+		$document->loadHTML( '<?xml encoding="utf-8" ?>' . $html );
 		libxml_clear_errors();
+
 		$document->encoding = 'UTF-8';
+
+		// Dirty fix to strip out existing XML Processing Instruction nodes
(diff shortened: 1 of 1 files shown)


src/Util/Util.php
```diff
@@ -185,10 +185,26 @@ public static function extractErrorMessage( ?ResponseInterface $resp, RequestInt
...
 		libxml_use_internal_errors( true );
-		$document->loadHTML( mb_convert_encoding( str_replace( '<?xml version="1.0" encoding="UTF-8" ?>', '', $html ), 'HTML-ENTITIES', 'UTF-8' ) );
+		// enforce a UTF-8 encoding declaration
+		$document->loadHTML( '<?xml encoding="utf-8" ?>' . $html );
 		libxml_clear_errors();
+
 		$document->encoding = 'UTF-8';
+
+		// Dirty fix to strip out existing XML Processing Instruction nodes
+		// (we already have one from the creation of the DOMDocument)
```
samwilson, line 197: Is it possible to `loadHTML()` without adding the PI? II feels a bit inefficient to loop through the whole document (which could be quite large) only to find a dupe of something we've just added here. (Sorry if I'm missing the obvious!)

Tpt:
Maybe dumb idea: instead of hacking around the PHP XML parser, what about using [RemexHTML](https://www.mediawiki.org/wiki/RemexHtml)? It is the HTML parser developped for Parsoid so we might hope it properly handles all theses cases.

inductiveload:
@Tpt sounds like a good idea, because whatever I do with the above, it seems to trip over on various inputs.

samwilson:
I think this is all sorted now in #479. Sorry we never resolved it years ago!

PR #376, opened 2021-09-14
Sanitize lang URL parameter

samwilson:
Restrict allowed characters in the lang parameter.

Bug: T290674


Daimona (approved):
I think this should eventually be rewritten for https://phabricator.wikimedia.org/T289155, but it's good for now.

samwilson:
Yes, good idea. I think one issue is that we might want to allow more lang values in the URL than we do in the list of projects (although, maybe that is only `beta`, in which case it can be handled separately). Also, there's still unresolved discussion around adding wikibooks support (e.g. T268851), which used to be done via a lang code such as `en-wikibooks`.

PR #377, opened 2021-09-16
Drops the HTML file splitting hack

Tpt:
Useful for old reading software like Adobe versions before 2009.

Should be ok now hopefully with moderns reading systems.

Bug: T261238


samwilson (approved):
Works in my testing!

PR #380, opened 2021-09-22
Add PHP 8.0 to CI

samwilson:


Daimona:
Bumped doctrine-migrations-bundle for doctrine/migrations#1104

Daimona:
I think the following PHP8 notice should be fixed:

>   1x: Required parameter $code follows optional parameter $prefix
>    1x in BookCreatorIntegrationTest::testCreateBookEpub3 from App\Tests\BookCreator

Then I see there are more hidden deprecation warnings, but those are also for PHP 7.3, hence not blocking this.

Daimona:
> > I think the following PHP8 notice should be fixed:
> > > 1x: Required parameter $code follows optional parameter $prefix
> > > 1x in BookCreatorIntegrationTest::testCreateBookEpub3 from App\Tests\BookCreator
> > 
> > 
> > Then I see there are more hidden deprecation warnings, but those are also for PHP 7.3, hence not blocking this.
> 
> That looks to be coming [from Intuition](https://github.com/Krinkle/intuition/blob/39cc5fbb3ced1d1bff166bbe3434b5f1953bdd06/language/mw-classes/Language.php#L3073) which in turn is getting it from MediaWiki, where it's already [been fixed](https://gerrit.wikimedia.org/g/mediawiki/core/+/4954ac1496b3656b6e9a9aae8ee896fd765f64a6/includes/language/Language.php#4383). I think Intuition needs to update it's mw-classes directory… I'll look into that.

Oh, I didn't realize it was from an external dependency! It might not be easy to fix then, considering that AFAICS those classes were last touched several years ago. Given that this change only adds testing for PHP 8 (and doesn't change the requirement in composer.json, which already includes PHP 8 BTW), and that this is just a notice, I'm going to merge it as-is.

PR #383, opened 2021-09-22
Bump phan, clean up config

Daimona:
Remove the baseline, since all issues were fixed. Also add a bunch of plugins based on normal MW config.


src/EpubCheck/Result.php
```diff
@@ -36,7 +36,7 @@ public function getLocations(): array {
 
 	public function toString(): string {
```
samwilson, line 37: It seems that this method isn't ever used so it's not surprising that this wrong method name never caused an issue!

This reminds me that there's still some refactoring to be done with the CheckCommand and the fact that it doesn't use these EpubCheck classes.

samwilson (approved):

PR #386, opened 2021-11-01
Add rate limiter to prevent bots and crawlers from pounding the site

MusikAnimal:
Bug: T286048


MusikAnimal:
Notes for reviewers:
- Set APP_RATE_LIMIT (number of requests) and APP_RATE_DURATION (number of minutes) as desired in your .env.local
- Comment out line 80 in RateLimitSubscriber. This should really check if the environment is prod, but it doesn't, for now :/

MusikAnimal:
I'm not sure why the build is failing. I thought it would copy over the values from `.env`. It seemed it didn't so I explicitly set them in `.env.test`, yet it's still complaining that they aren't set?

.env.test
```diff
@@ -1 +1,3 @@
 SYMFONY_DEPRECATIONS_HELPER=quiet[]=indirect&quiet[]=other
+APP_RATE_LIMIT=0
+APP_RATE_DURATION=0
```
samwilson, line 3: These two shouldn't be required here.

config/services.yaml
```diff
@@ -65,5 +65,10 @@ services:
         arguments:
             $projectDir: '%kernel.project_dir%'
 
+    App\EventSubscriber\RateLimitSubscriber:
+        arguments:
+            $rateLimit: '%env(int:APP_RATE_LIMIT)%'
+            $rateDuration: '%env(int:APP_RATE_DURATION)%'
+
```
samwilson, line 72: This section needs to be specified in `config/services_test.yaml` as well. I'm pretty sure there's a way to share these, but it's not set up at the moment. :(

This is where the CI failure is coming from.

src/EventSubscriber/RateLimitSubscriber.php
```diff
@@ -0,0 +1,111 @@
+<?php
+declare( strict_types=1 );
+
+namespace App\EventSubscriber;
+
+use DateInterval;
+use Krinkle\Intuition\Intuition;
+use Psr\Cache\CacheItemPoolInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
```
samwilson, line 9: Unused use statement.

src/EventSubscriber/RateLimitSubscriber.php
```diff
@@ -0,0 +1,111 @@
...
+			[ , $action ] = $controller;
+		}
+
+		// Abort if rate limitations are disabled or we're not exporting a book.
+		if ( $this->rateLimit + $this->rateDuration === 0 || $action !== 'home' || !$request->get( 'page' ) ) {
+			return;
+		}
+
+		$xff = $request->headers->get( 'x-forwarded-for', '' );
+		if ( '' === $xff ) {
```
samwilson, line 78: Not allowed is the use of Yoda conditions.

samwilson (commented):
This looks good, and I think works as much as I've tested it.

We talked at some point about adding this to the Toolforge bundle; do you think it's worth it? Or are there aspects that would be annoying to have on every tool?

MusikAnimal:
> We talked at some point about adding this to the Toolforge bundle; do you think it's worth it? Or are there aspects that would be annoying to have on every tool?

Yes, perhaps! My only concern is that this relies on the X-Forwarded-For header being enabled, which as I understand it is only enabled under rare circumstances and by special permission from Cloud Services. I think XTools and WS Export are the only two tools (under our umbrella, at least) that would benefit from this. But of course with the default values for rate limiting set to zero, it won't hurt any other clients to have this code in the bundle.

src/EventSubscriber/RateLimitSubscriber.php
MusikAnimal: Done!

config/services.yaml
MusikAnimal: Ah, that explains it. Done!

src/EventSubscriber/RateLimitSubscriber.php
samwilson: I think it would've been caught by the `MediaWiki.Classes.UnusedUseStatement.UnusedUse` sniff, which we're already using. It didn't only because CI didn't get that far, I think.

samwilson:
> > We talked at some point about adding this to the Toolforge bundle; do you think it's worth it? Or are there aspects that would be annoying to have on every tool?
> 
> Yes, perhaps! My only concern is that this relies on the X-Forwarded-For header being enabled, which as I understand it is only enabled under rare circumstances and by special permission from Cloud Services. I think XTools and WS Export are the only two tools (under our umbrella, at least) that would benefit from this. But of course with the default values for rate limiting set to zero, it won't hurt any other clients to have this code in the bundle.
> 
> I'll also add that the [XTools implementation](https://github.com/x-tools/xtools/blob/ed153bf24c10f8d6b31d12e93d47cf7a96a93bd4/src/AppBundle/EventSubscriber/RateLimitSubscriber.php) is more complex. The code introduced with this PR isn't enough to suit its needs, but we could easily add a second EventSubscriber class to do the extra things, if we wanted (though XTools isn't using the bundle yet anyway).

Yes, that all makes total sense. Best to keep it here (and in XTools) then. Maybe one day a whole-toolforge fix will be made for this problem!

drzraf:
Do you mind relaxing a bit the values unless the server is **really** on its knees?

It's become impossible to simply export a file in PDF (which is a pure read-only operation) from Wikisource Book Export when using a VPN. (Caching exported PDF files would lower server load and would make more sense than caching IP address to rate-limit access ;))

drzraf:
NB: https://phabricator.wikimedia.org/T286048 view is restricted

PR #387, opened 2021-11-04
Add logging for rate-limiter, and fix small CSS bug

samwilson:
* Add some debugging for the rate-limiting, to help figure out what's going on.
* Fix the display of the error message icon (noticed because the rate-limiting
  message is more than one line).

Bug: T286048


samwilson:
This is probably not needed, actually.

samwilson:
I've moved the CSS fix to its own PR: #388

PR #388, opened 2021-11-04
Add space around alert icon

samwilson:


Daimona:
Sorry, I'm still not very familiar with the codebase. How could I test this easily?

samwilson:
> Sorry, I'm still not very familiar with the codebase. How could I test this easily?

Lots of the [installation](https://github.com/wikimedia/ws-export#installation) can be skipped, for this bug: basically, get the code, run `composer install` and `symfony server:start`, and it should be possible to load the main export form. Give it a non-existing page title, and it'll give an error message showing the mis-aligned 'X' icon (normally, exporting will want other things like Calibre and a database to save stats to, but they don't matter here).

samwilson:
Oh, and the bug can also be seen at e.g. https://ws-export-test.wmcloud.org/?page=foobar

Daimona:
Ah, thanks, LGTM!

PR #390, opened 2021-12-13
Don't link to non-existent wiki lang

aezell:
We already know that we don't want to link to a lang if it's not set, so this just makes it more specific. It must be set and must exist in the area of existing wiki langs.

### DIFF
--- a/src/Controller/ExportController.php
+++ b/src/Controller/ExportController.php
@@ -118,6 +118,7 @@ class ExportController extends AbstractController {
 			'formats' => GeneratorSelector::getValidFormats(),
 			'format' => $this->getFormat( $request ),
 			'title' => $this->getTitle( $request ),
+			'wikiLangs' => $this->wikidata->getWikisourceLangs( $this->intuition->getLang() );
 			'langs' => $this->getLangs( $request ),
 			'lang' => $this->getLang( $request ),
 			'credits' => $credits,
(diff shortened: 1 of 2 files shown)


aezell:
Now that I look closer, it looks like `getLangs()` is always going to add the one from the URL to the array passed to the template. So, I guess what I need is another function that only gets the actual wiki langs and passes that into the function. Something like `getWikiLangs()`?

aezell:
I did it that way. New array into the template of just the existing wikis. The lang logic then lives solely in the template.

src/Controller/ExportController.php
```diff
@@ -118,6 +118,7 @@ public function home(
 			'formats' => GeneratorSelector::getValidFormats(),
 			'format' => $this->getFormat( $request ),
 			'title' => $this->getTitle( $request ),
+			'wikiLangs' => $this->wikidata->getWikisourceLangs( $this->intuition->getLang() ),
```
samwilson, line 121: The `getWikisourceLangs()` method gives an array of lang codes to names and the `lang` twig var is always a code, so either this needs to be wrapped with `array_keys()`, or the check below needs to be something like `{% if wikiLangs[lang] is defined %}`.

samwilson (commented):
Hello Alex! :-)

This looks good, I think it just needs a tiny tweak.

MusikAnimal:
So great to see you here, Alex!

src/Controller/ExportController.php
aezell: Or, we could use the `|keys` filter maybe like `{% if lang in wikiLangs|keys %}`?

aezell: I added a commit that does just that.

samwilson: Yes, that's neater, good idea.

samwilson (approved):

PR #394, opened 2022-02-01
Deny access from IPs globally blocked on Wikimedia wikis

MusikAnimal:
This is to prevent automated disruption, since stewards are pretty on
top of blocking webhosts. Some users with a VPN may get blocked, but
will be able to login to get around it (as well as the preexisting
IP-based rate limiting).

A 'Login' link is not shown in the normal UI, only on the error page,
because this is the only situation where you'd need to login.

Bug: T286048


MusikAnimal:
I went to set this up on staging then realized we don't have the XFF headers enabled there... requested that at https://phabricator.wikimedia.org/T279111#7669654

There's no easy way to test this locally either since it looks at the XFF header, not your actual IP. You could put in an IP that is known to be blocked as the default value at https://github.com/wikimedia/ws-export/pull/394/files#diff-ea4028fef53c949e95765094ba8b64d33486feda9281ee8dcc7e0dfaff139136R90 like so:
```
$xff = $request->headers->get( 'x-forwarded-for', '94.119.64.73' );
```
but this still won't let you test the error page, unless you get your local to run in production mode.

Cloud Services is usually semi-quick so I imagine they'll get the XFF headers added to staging within a day or two.

samwilson:
This looks good to me, although I've not been able to definitively test it.

PR #397, opened 2022-02-13
Atom: Makes sure to escape '&'

Tpt:
DOMDocument::createElement escapes '<' and '>' but not '&'...

### DIFF
--- a/src/Generator/AtomGenerator.php
+++ b/src/Generator/AtomGenerator.php
@@ -120,7 +120,7 @@ class AtomGenerator implements FormatGenerator {
 			return;
 		}
 
-		$node = $dom->createElement( $name, $value );
+		$node = $dom->createElement( $name, str_replace( '&', '&amp;', $value ) );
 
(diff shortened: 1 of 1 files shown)


src/Generator/AtomGenerator.php
```diff
@@ -120,7 +120,7 @@ private function addNode( DOMDocument $dom, DOMElement $head, $name, $value, $ty
 			return;
 		}
 
-		$node = $dom->createElement( $name, $value );
+		$node = $dom->createElement( $name, str_replace( '&', '&amp;', $value ) );
```
samwilson, line 123: Actually, I think another way to do this might be to set `textContent` directly, e.g.:

```php
		$node = $dom->createElement( $name );
		$node->textContent = $value;
```

I'm not sure if there are other considerations, so probably it's fine to handle the one-off case you've found.

Tpt: It sounds much better. Thank you!  I create #398 to do it.

PR #400, opened 2022-03-24
Add 'CLI Usage' section to README.md

samwilson:
Auto-generate the CLI usage docs from the `app:` namespace Commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,8 +55,47 @@ $ ./php bin/console toolforge:ssh
 $ php bin/console toolforge:ssh --bind-address=0.0.0.0
 ```
 
-Tests
-=====
+## CLI Usage
+
+### app:check
+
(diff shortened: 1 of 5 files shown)


MusikAnimal (approved):
Works beautifully! This is a very nifty tool! I was trying to find it last night... I thought there was some sort of Symfony "marketplace" (or maybe it was Flex, which is now going [server-less](https://symfony.com/blog/symfony-flex-is-going-serverless)); anyway what I'm saying this should be promoted somehow within the Symfony community :)

samwilson:
Yes, I keep seeing warnings about Flex! We'll have a bit of upgrading to do I guess.

And I'm not sure where the best place to announce Symfony projects is. https://symfony.com/community says maybe Slack, so maybe I'll try that. There's a [couple of things](https://github.com/samwilson/console-readme-generator/issues) I want to work on first I think (and can you think of any more?).

PR #401, opened 2022-03-24
Add missing author names to toolinfo.json

samwilson:
These names were the result of the following command:

git log --format='{"name": "%aN"},' | sort |  uniq -c | sort -nr


samwilson:
This is because Toolhub now supports multiple author names.

Tpt (approved):

PR #408, opened 2022-05-06
Remove OSTYPE check and update README file

cimurah:
Remove OSTYPE check in Dockerfile since it is not
necessary. Its value is empty when image is being built.

Cleaned up commands in README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,11 +48,11 @@ Installation
 
   * Establish an SSH tunnel to the replicas (only necessary on local environments)
 ```bash
-$ ./php bin/console toolforge:ssh
+$ ./bin/console toolforge:ssh
 ```
   * Bind address for docker enviroments
 ```bash
-$ php bin/console toolforge:ssh --bind-address=0.0.0.0
+$ ./bin/console toolforge:ssh --bind-address=0.0.0.0
 ```
 
 ## CLI Usage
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -10,7 +10,7 @@ EXPOSE 8080
 # work around https://github.com/docker-library/openjdk/blob/0584b2804ed12dca7c5e264b5fc55fc07a3ac148/8-jre/slim/Dockerfile#L51-L54
 RUN mkdir -p /usr/share/man/man1
 
-RUN if [[ "$OSTYPE" == "linux-gnu"* ]]; then addgroup --gid $GROUP_ID docker; fi
+RUN addgroup --gid $GROUP_ID docker
```
samwilson, line 13: I don't have Windows so can't test this, but I assume it's all good!

README.md
```diff
@@ -48,7 +48,7 @@ Installation
 
   * Establish an SSH tunnel to the replicas (only necessary on local environments)
 ```bash
-$ ./php bin/console toolforge:ssh
+$ php bin/console toolforge:ssh
```
samwilson, line 51: This should probably match the below and be just `./bin/console …`

samwilson (approved):
Looks good to me.

README.md
cimurah: Done. Thank you!

PR #412, opened 2022-06-21
Update PHP dependencies

samwilson:


theresnotime (approved):

MusikAnimal:
The deletion of the `symfony.lock` file should be committed, otherwise the repo is dirty after pulling in these changes and running `composer install`.

PR #450, opened 2023-03-15
Add Parsoid alignment styles to default stylesheet

samwilson:
Bug: T330949

### DIFF
--- a/resources/styles/mediawiki.css
+++ b/resources/styles/mediawiki.css
@@ -47,3 +47,61 @@ div.tleft {
 .small {
 	font-size: 94%;
 }
+
+/* Styles simplified from resources/src/mediawiki.skinning/content.media-common.less */
+.mw-valign-middle {
+	vertical-align: middle;
+}
+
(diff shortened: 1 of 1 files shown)


Tpt:
Thank you! Genuine question: is there a way to fetch it automatically from MediaWiki to avoid having to synchronize them again in the future? If yes, is it a good idea?

samwilson:
I initially thought that would be the way to do this, yes, but actually I suspect that ereaders are less forgiving of complex CSS than web browsers are, so perhaps it's better that we do it manually. I'm not really sure though! :)

Tpt:
> but actually I suspect that ereaders are less forgiving of complex CSS than web browsers are, so perhaps it's better that we do it manually. I'm not really sure though! :)

Yes, it makes a lot of sense. Thank you!

samwilson:
@Tpt the failures here are fixed in #451.

Tpt (approved):
@samwilson Thank you!

PR #468, opened 2023-07-13
Remove closed Wikisources from language dropdown

Poslovitch:
Modify the query to ignore Wikisources that are instances of "closed wiki".

Bug: https://phabricator.wikimedia.org/T276887


samwilson:
Thanks!

PR #470, opened 2023-07-24
Hard-code check for Multilingual Wikisource to avoid errors

samwilson:
Rather than looking for an empty subdomain, check for the Wikidata ID of Multilingual Wikisource.

Bug: T342520

### DIFF
--- a/src/Wikidata.php
+++ b/src/Wikidata.php
@@ -37,7 +37,7 @@ class Wikidata {
 			$cacheItem->expiresAfter( new DateInterval( 'P1M' ) );
 			$this->logger->debug( "Requesting list of Wikisources from Wikidata" );
 			$query =
-				"SELECT ?label ?website WHERE { "
+				"SELECT ?item ?label ?website WHERE { "
 				// Instance of Wikisource language edition but not of closed wiki.
 				. "?item wdt:P31 wd:Q15156455 . "
 				. "MINUS { ?item wdt:P31 wd:Q47495990 . } "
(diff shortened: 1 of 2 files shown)


src/Wikidata.php
```diff
@@ -54,12 +54,14 @@ public function getWikisourceLangs( string $lang ): array {
...
-					$subdomain = 'mul';
+				// Hard-code Multilingual Wikisource, to avoid issues with incubator Wikisources
+				// being given the same domain name as P856 (official website). T342520.
+				if ( str_ends_with( $datum['item'], 'Q18198097' ) ) {
+					$out['mul'] = $datum['label'];
+					continue;
 				}
-				$out[$subdomain] = $datum['label'];
+				preg_match( '|https://([a-z-_]*)\.?wikisource\.org|', $datum['website'], $matches );
+				$out[$matches[1]] = $datum['label'];
```
Tpt, line 64: `$matches[1]` seems a bit fragile to me if Wikidata data are invalid for some reason.

Tpt (approved):

src/Wikidata.php
samwilson: Good point. I did initially try it with *Wikimedia language code* (P424), but that doesn't get the correct value for some, like syl. Although maybe https://www.wikidata.org/wiki/Q120844812 should not be called a Wikisource language edition yet, while it's in incubator.

If someone puts an invalid value for Official website then you're right, this would break, but then whatever we pull from Wikidata could get broken at some point so I'm not sure it's too different.

Tpt: Sorry, I have made this disucssion slip. Indeed, that's a good point. I am fine with merging this PR.

samwilson:
This has been sorted out on Wikidata, no changes needed here now I think.

PR #474, opened 2023-08-19
Improve and expand Util::encodeString

waldyrious:
- Expand support for digraphs
- Ruplicate encoding for ψ as both 'c' and 'ps'
- Expand support for Latin diacritics in Util::encodeString
- Use a data provider for the `Util::encodeString` tests
- Don't lowercase non-Latin characters in `Util::encodeString`
  Also add missing variant of lowercase Greek sigma
- Add a Greek pangram test case for `Util::encodeString`
- Add support for Cyrillic in `Util::encodeString`
- Expand coverage of `Util::encodeString` to many more scripts
  Using [`transliterator_transliterate`](https://www.php.net/manual/en/transliterator.transliterate.php), we are able to cover many more scripts,
  without resorting to hardcoded replacements.
  This yields slightly different results,
  but greatly expands the coverage of scripts that result in readable filenames.

These changes have been submitted in separate atomic commits.
I recommend reviewing them separately,
and kindly request the merge to be done without squashing;
otherwise the result will be a large amorphous diff where it's hard to parse what actually happened.


samwilson (approved):

samwilson:
Thanks! Looks good.

PR #478, opened 2023-11-30
Add libxcb-cursor0 for Calibre

samwilson:
For installing Calibre in CI, and getting tests to pass.


Tpt (approved):

samwilson:
@tpt thanks for looking at these so quickly! :-) I've just changed this one to just be about Calibre (I was not running the right vendor dir locally before).

PR #479, opened 2023-11-30
Simplify loading of HTML

samwilson:
This avoids the deprecated mb_convert_encoding usage, as well as
the encoding of characters in non-HTML contexts.

Bug: T271390

### DIFF
--- a/src/Util/Util.php
+++ b/src/Util/Util.php
@@ -74,7 +74,7 @@ class Util {
 		if ( $lang != null ) {
 			$html .= ' xml:lang="' . $lang . '" dir="' . static::getLanguageDirection( $lang ) . '"';
 		}
-		$html .= '><head><meta content="application/xhtml+xml;charset=UTF-8" http-equiv="default-style" /><link type="text/css" rel="stylesheet" href="main.css" /><title>' . $title . '</title></head>';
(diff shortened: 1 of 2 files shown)


src/Util/Util.php
```diff
@@ -183,7 +183,7 @@ public static function extractErrorMessage( ?ResponseInterface $resp, RequestInt
 	public static function buildDOMDocumentFromHtml( string $html ): DOMDocument {
 		$document = new DOMDocument( '1.0', 'UTF-8' );
 		libxml_use_internal_errors( true );
-		$document->loadHTML( mb_convert_encoding( str_replace( '<?xml version="1.0" encoding="UTF-8" ?>', '', $html ), 'HTML-ENTITIES', 'UTF-8' ) );
+		$document->loadHTML( str_replace( '<?xml version="1.0" encoding="UTF-8" ?>', '', $html ), LIBXML_HTML_NODEFDTD );
```
Tpt, line 186: Out of curiosity question: I have seen you tried using the parser from `wikimedia/html-formatter`. Why is it not working?

Tpt (approved):

src/Util/Util.php
samwilson: Yeah, I tried that, but it produces HTML rather than XHTML so validation kept failing for various reasons. :( I might have been doing something wrong though!

Tpt: Thank you!

PR #494, opened 2023-12-20
Redesign stats page

Parthiv-M:
Redesign stats page to be cleaner and
convey stats in a better way with more
accessibility for year/month inputs.
Also added i18n for content on the page.

Bug: [T353737](https://phabricator.wikimedia.org/T353737)

### DIFF
--- a/templates/statistics.html.twig
+++ b/templates/statistics.html.twig
@@ -41,7 +41,7 @@
 		<div class="col-md-8">
 			{% include '_month_year_selector.html.twig' with {month: month, year: year} %}
 			<table class="table table-striped">
-				<caption>{{ msg('stat-date-label') }} {{ month }}/{{ year }}</caption>
+				<caption>Stats for {{ month }}/{{ year }}</caption>
 				<thead>
 					<tr>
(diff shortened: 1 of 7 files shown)


templates/statistics.html.twig
```diff
@@ -2,76 +2,83 @@
...
+		<li class="active">{{ msg('breadcrumbs-stats-link') }}</li>
+	</ol>
+	<div class="row stats">
+		<div class="col-md-4 recents">
+			<h1 class=""><strong>{{ msg('recently-popular-heading') }}</strong></h1>
+			<h4 class="text-muted">{{ msg('recently-popular-subtext') }}</h4>
+			<ol>
+				{% for book in recently_popular %}
+					<li class="recent row">
+						<div class="col-md-9">
```
samwilson, line 17: This col and the one after it could perhaps be `col-xs-` ones so that the side-by-side alignment is kept on narrow screens. Although perhaps that's not great for long book titles. Anyway, can be done later! :)

samwilson (approved):

PR #495, opened 2023-12-23
Add iiurlwidth API param when fetching cover image info

samwilson:
To avoid errors when working with TIFF cover images.

Bug: T353954


Tpt (approved):
Thank you!

PR #496, opened 2023-12-28
Make only one API request for cover image info

samwilson:
To avoid a 2nd API request, always use 400px for the image width, and use the file extension to determine if the file is PDF/DjVu.

### DIFF
--- a/src/BookProvider.php
+++ b/src/BookProvider.php
@@ -267,8 +267,8 @@ class BookProvider {
 	}
 
 	/**
-	 * return the cover of the book
-	 * @param $cover string the name of the cover
+	 * Return the cover of the book.
+	 * @param string $cover The title of the cover without 'File', e.g. Lorem.pdf/3
 	 * @return ?Picture The cover picture, or null if it could not be determined.
 	 */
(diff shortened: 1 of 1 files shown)


samwilson:
I'm not 100% sure this is the right way to go.

* Are there times when we want to use an image smaller than 400px?
* It looks like we don't need to limit to the max page number (via pagecount) because if we request a number higher than the total number of pages Thumbor resolves it to ([e.g. requesting page 999 results in 176](https://commons.wikimedia.org/wiki/Special:ApiSandbox#action=query&format=json&prop=imageinfo&titles=file%3APaquin%20-%20Jules%20Faubert%2C%20le%20roi%20du%20papier%2C%201923.djvu&formatversion=2&iiprop=mime%7Cdimensions%7Curl&iiurlwidth=400&iiurlparam=page999-400px)).

Tpt (approved):
> Are there times when we want to use an image smaller than 400px?

I think only if the original image is smaller than 400px. But it sounds not very likely and it won't increase the epub size significantly anyway.

> It looks like we don't need to limit to the max page number (via pagecount)

Great!

PR #497, opened 2023-12-30
Uses a UNIX semaphore to limit concurrent ebook-convert calls

Tpt:
Sadly these semaphores don't work on Windows

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ This section deals with a full installation of WS Export on the local developmen
    For example, in Debian-based Linux distributions:
 
    ```console
-   apt install php-sqlite3 php-zip php-curl
+   apt install php-sqlite3 php-zip php-curl php-sysvsem
    ```
 
    Then create a `.env.local` file:
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -70,5 +70,11 @@ services:
             $rateLimit: '%env(int:APP_RATE_LIMIT)%'
             $rateDuration: '%env(int:APP_RATE_DURATION)%'
 
+    App\Util\Semaphore\Semaphore:
+        class: 'App\Util\Semaphore\UnixSemaphore'
+        arguments:
+            $semaphoreKey: 123455435644
+            $capacity: 4
+
     # add more service definitions when explicit configuration is needed
     # please note that last definitions always *replace* previous ones
--- a/src/Generator/ConvertGenerator.php
+++ b/src/Generator/ConvertGenerator.php
@@ -5,6 +5,7 @@ namespace App\Generator;
 use App\Book;
 use App\Exception\WsExportException;
 use App\FileCache;
+use App\Util\Semaphore\Semaphore;
 use App\Util\Util;
 use InvalidArgumentException;
 use Symfony\Component\HttpFoundation\Response;
@@ -77,9 +78,7 @@ class ConvertGenerator implements FormatGenerator {
 		return array_keys( self::$CONFIG );
 	}
 
-	/**
-	 * @var string
-	 */
+	/** @var string */
 	private $format;
 
 	/** @var int Command timeout in seconds. */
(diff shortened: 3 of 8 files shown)


config/services.yaml
```diff
@@ -70,5 +70,11 @@ services:
             $rateLimit: '%env(int:APP_RATE_LIMIT)%'
             $rateDuration: '%env(int:APP_RATE_DURATION)%'
 
+    App\Util\Semaphore\Semaphore: '@App\Util\Semaphore\UnixSemaphore'
+    App\Util\Semaphore\UnixSemaphore:
+        arguments:
+            $semaphoreKey: 123455435644
+            $capacity: 4
```
Tpt, line 77: I am not 100% sure this is configured properly.

src/Generator/ConvertGenerator.php
```diff
@@ -91,10 +90,14 @@ public static function getSupportedTypes() {
...
 
-	public function __construct( int $timeout, FileCache $fileCache, EpubGenerator $epubGenerator ) {
+	/** @var ?Semaphore */
+	private $semaphore;
+
+	public function __construct( int $timeout, FileCache $fileCache, EpubGenerator $epubGenerator, ?Semaphore $semaphore = null ) {
 		$this->timeout = $timeout;
 		$this->fileCache = $fileCache;
 		$this->epubGenerator = $epubGenerator;
+		$this->semaphore = $semaphore;
```
samwilson, line 100: I think there might be a problem with injecting a Semaphore here in that it'll be instantiated for every request (including those that don't use ConvertGenerator, e.g. just opening the main export form). Then `sem_get()` will be called every time, and won't that mean that it'll run out of lock acquisitions? An alternative might be to inject a SemaphoreFactory or similar, or move `sem_get()` into `tryLock()`?

README.md
```diff
@@ -40,7 +40,7 @@ This section deals with a full installation of WS Export on the local developmen
    For example, in Debian-based Linux distributions:
 
    ```console
-   apt install php-sqlite3 php-zip php-curl
+   apt install php-sqlite3 php-zip php-curl php-sysvsem
```
samwilson, line 43: Should also be added as a requirement in `composer.json`.

Tpt: This means that installation would break on Windows (no sysvsem module). Sadly it does not seems there is a way to set optional dependencies in composer.

src/Generator/ConvertGenerator.php
Tpt: `sem_get` only creates the semaphore if it does not exists yet and returns its resource descriptor but does not lock it. Locking is done with `sem_acquire`. I have updated the MR to lazily call `sem_get` only on `tryLock()` to avoid running it if the semaphore is not used. Thank you!

samwilson: Ah makes sense, thanks! And looks good now.

src/Generator/ConvertGenerator.php
```diff
@@ -91,10 +90,14 @@ public static function getSupportedTypes() {
 	/** @var EpubGenerator */
 	private $epubGenerator;
 
-	public function __construct( int $timeout, FileCache $fileCache, EpubGenerator $epubGenerator ) {
+	/** @var ?Semaphore */
+	private $semaphore;
+
+	public function __construct( int $timeout, FileCache $fileCache, EpubGenerator $epubGenerator, ?Semaphore $semaphore = null ) {
```
samwilson, line 96: Is `$semaphore` able to be null here in order to make the tests easier? Because it looks like in non-test code it never will be.

samwilson (approved):
Looks great! This all works in my tests. I'll get it merged and deployed now.

Is there an issue with running the tests if sysvsem isn't installed/enabled? I'm not sure if that matters too much right now, as we don't seem to have any developers working on Windows.

src/Generator/ConvertGenerator.php
Tpt: Yes, and to allow people on Windows to bypass the need to provide a semaphore.

PR #505, opened 2024-02-26
Limit the concurrency when fetching pages

tstarling:
For a production test case, BookProvider::getPages() was opening 3492
concurrent connections to the Wikimedia servers. That exceeds the
default maxmimum file descriptor limit, and there is a risk that the
tool will be blocked by WMF sysadmins.

So, limit the concurrency to 10 connections.


Tpt (approved):
Thank you so much! I am learning quite a lot of async PHP here.

PR #509, opened 2024-03-11
Remove old book.php URLs

samwilson:
They're now handled alonside other legacy URLs in the web server config, documented at https://wikitech.wikimedia.org/wiki/Nova_Resource:Wikisource/Wikisource_Export

Bug: T359659


samwilson:
This is not needed. Fixed the Apache config instead.

PR #510, opened 2024-03-19
Fix issue with cleaning transcluded reference links

samwilson:
This fixes two issues with the PageParser::cleanReferenceLinks() method: the first when there are no reference links found, the second the actual bug linked below. The `typeof` attribute can contain values other than 'mw:Extension/ref' but we were only checking for that one on its own. Switching to xPath contains() function fixes this.

Bug: T358965


Tpt (approved):
Thank you!

PR #512, opened 2024-07-17
Return early for cover images that are not found

samwilson:
BookProvider::getCover() was already returning for no pages returned, so this switches to API formatversion 2 and also checks for the `missing` parameter in the first/only returned page.

Bug: T370257


samwilson:
I think I might merge this if no one minds? Just to clear up the log spam if nothing else (and it'll help a few dozen people who daily try to download books that have incorrect cover images).

There are other related issues, but they seem mostly to be needed to be resolved on wikis. e.g. eswikisource are encoding their cover image; I left them a message at https://es.wikisource.org/wiki/Plantilla_discusi%C3%B3n:Proofreadpage_index_template#ws-cover

Tpt (approved):

Tpt:
Yes!

PR #516, opened 2024-12-22
Change the cover image to Wikisource (no-text) logo

harsv689:
The cover page image of the downloaded pdfs are Wikisource(no-text) logo now.


samwilson:
Thanks for working on this!

It looks like this is a work-in-progress, is that right?

harsv689:
Yes

samwilson:
This was done in #522.

PR #518, opened 2025-01-18
Upgrade to Symfony 7

samwilson:
Upgrade to Symfony 7 and update (almost) all PHP dependencies, making as few changes as possible.

Bug: T384115

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
 
     services:
       mysql:
-        image: mysql:5.7
+        image: mysql:5.8
         env:
           MYSQL_DATABASE: wsexport_test
           MYSQL_ROOT_PASSWORD: testpwd
@@ -30,7 +30,7 @@ jobs:
 
     env:
       APP_ENV: test
-      DATABASE_URL: mysql://root:testpwd@127.0.0.1:3306/wsexport_test?serverVersion=5.7
+      DATABASE_URL: mysql://root:testpwd@127.0.0.1:3306/wsexport_test?serverVersion=5.8
       EPUBCHECK_JAR: /home/runner/work/tool/tool/epubcheck-${{matrix.epubcheck}}/epubcheck.jar
 
     runs-on: ${{matrix.os}}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest ]
-        php: [ '7.3', '7.4', '8.0', '8.1', '8.2' ]
+        php: [ '8.2', '8.2', '8.4' ]
         epubcheck: [ '5.1.0' ]
 
     env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
       matrix:
         os: [ ubuntu-latest ]
         # Test against the lowest and highest PHP versions that we support.
-        php: [ '8.2', '8.4' ]
+        php: [ '8.2', '8.3' ]
         epubcheck: [ '5.1.0' ]
 
     env:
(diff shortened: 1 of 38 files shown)


.github/workflows/ci.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest ]
-        php: [ '7.3', '7.4', '8.0', '8.1', '8.2' ]
+        php: [ '8.2', '8.2', '8.4' ]
```
Tpt, line 27:
```suggestion
        php: [ '8.2', '8.3', '8.4' ]
```

Tpt: or maybe we don't care much about 8.3 if we already cover 8.2 and 8.4?

samwilson: Good point! I've updated it to 8.2 and 8.4 — you're right, 8.3 issues will likely be caught enough by the latter.

samwilson:
Another fix is needed to the ToolforgeBundle before this can be morged: https://github.com/wikimedia/ToolforgeBundle/pull/71

samwilson:
Okay, this is ready for review now, if anyone's got time.

.github/workflows/ci.yml
```diff
@@ -24,12 +24,13 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest ]
-        php: [ '7.3', '7.4', '8.0', '8.1', '8.2' ]
+        # Test against the lowest and highest PHP versions that we support.
+        php: [ '8.2', '8.3' ]
```
Tpt, line 28: I guess PHP 8.4 is broken?

src/Util/Semaphore/UnixSemaphore.php
```diff
@@ -13,7 +14,7 @@ class UnixSemaphore implements Semaphore {
 	private $semaphoreKey;
 	/** @var int */
 	private $capacity;
-	/** @var resource|null the lazily initialized semaphore descriptor */
+	/** @var resource|SysvSemaphore|null the lazily initialized semaphore descriptor */
```
Tpt, line 17: genuine question: do we still need `resource` now that we have `SysvSemaphore`?

(same in other places)

tests/Repository/GeneratedBookRepositoryTest.php
```diff
@@ -22,17 +22,18 @@ class GeneratedBookRepositoryTest extends KernelTestCase {
...
 	}
 
 	/**
-	 * @covers GeneratedBookRepository::getTypeAndLangStats
+	 * @covers \App\Repository\GeneratedBookRepository::getTypeAndLangStats
 	 */
 	public function testAddAndRetrieve() {
 		// Make sure it's empty to start with.
-		static::assertEmpty(
+		static::assertSame(
```
Tpt, line 35: for me: Is there a difference between `static::assertSame` and `$this->assertSame`? I would guess they are the same function and we should harmonize on something?

Tpt (approved):

tests/Repository/GeneratedBookRepositoryTest.php
samwilson: Good point. [The docs](https://docs.phpunit.de/en/11.5/assertions.html#appendixes-assertions) say that there's no right way, and as we've got more `$this->` than `static` I'll stick to that! And also use `assertSame` over `assertEquals` unless we want loose checking (which it doesn't look like we do anywhere).

src/Util/Semaphore/UnixSemaphore.php
samwilson: Good point, no it seems not. Removed!

.github/workflows/ci.yml
samwilson: Yeah it kept failing (and fixing it would make it fail for 8.2), and as we're not going to be deploying to PHP 8.4 any time soon I figured it could wait.

PR #520, opened 2025-01-29
Update epubcheck to v5.2.1

samwilson:
Also check that the file exists so that it fails if epubcheck is not installed.


Tpt (approved):

samwilson:
I guess epubcheck has been failing for ages, since we changed the repo name! But not failing visibly.

PR #531, opened 2025-02-14
Update NPM dependencies and remove Gruntfile.js

samwilson:
Move the two commands in the Gruntfile into package.json, as many other Wikimedia projects are doing these days.


Parthiv-M:
Can confirm that the script works fine.

PR #532, opened 2025-02-16
Add language switcher

samwilson:
Add a language-switching dropdown menu in the site footer.

Bug: T387426

### DIFF
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -83,5 +83,5 @@
 	"breadcrumbs-stats-link": "Link of statistics page shown in the breadcrumb",
 	"recently-popular-heading": "Heading text for recently popular section",
 	"recently-popular-subtext": "Description for recently popular section",
-	"change-lang": "Button label for the footer form with which to switch between UI languages."
+	"change-lang": "Label for the button in the footer that allows switching between UI languages."
 }
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -82,5 +82,6 @@
 	"breadcrumbs-home-link": "Link of home page shown in the breadcrumb",
 	"breadcrumbs-stats-link": "Link of statistics page shown in the breadcrumb",
 	"recently-popular-heading": "Heading text for recently popular section",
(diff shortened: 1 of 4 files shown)


i18n/qqq.json
```diff
@@ -82,5 +82,6 @@
 	"breadcrumbs-home-link": "Link of home page shown in the breadcrumb",
 	"breadcrumbs-stats-link": "Link of statistics page shown in the breadcrumb",
 	"recently-popular-heading": "Heading text for recently popular section",
-	"recently-popular-subtext": "Description for recently popular section"
+	"recently-popular-subtext": "Description for recently popular section",
+	"change-lang": "Button label for the footer form with which to switch between UI languages."
```
waldyrious, line 86:
```suggestion
	"change-lang": "Label for the button in the footer form that allows switching between UI languages."
```

waldyrious:
A screenshot would be nice :pray:

samwilson:
Good point! This should also have a Phab task, for visibility. I've created [T387426](https://phabricator.wikimedia.org/T387426).

i18n/qqq.json
samwilson: That sounds better, agreed.

templates/base.html.twig
```diff
@@ -75,6 +75,22 @@
...
 			</p>
+			<form method="get" class="form-inline change-lang">
+				<p>
+					<select name="uselang" class="form-control">
+						{% for code,lang in all_langs() %}
+							<option value="{{code}}" {% if code == lang() %}selected{% endif %}>
+								{% if lang %}
+									{{lang}}
+								{% else %}
+									({{code}})
```
samwilson, line 86: This fallback shouldn't really be necessary, but not all languages have names in Intuition yet. (I think [T331584](https://phabricator.wikimedia.org/T331584) is the task for that, and perhaps `all_langs()` should handle the fallback for us; that's part of the [ToolforgeBundle](https://github.com/wikimedia/ToolforgeBundle).)

waldyrious (approved):
LGTM!

PR #533, opened 2025-02-17
Inject bookprovider

samwilson:


samwilson:
Ignore this. I'll come back to it later when I'm clearer about the next step.

PR #536, opened 2025-02-27
Update PHP dependencies

samwilson:
All are minor updates.


samwilson:
Self-merging as these are all minor updates and all tests pass.

